PR #12, opened 2016-05-19
Add omp critical section to PlmBar

opoplawski:
I've been starting to port some SHTOOLS using code to use OpenMP.  I had to add this section to PlmBar() to avoid multiple runs of the init code.  It seems to work reasonably well for me.  I've not looked at any other of the SHTOOLS routines.

Locally we are now building a separate copy of SHTOOLS (libSHTOOLS_thread.a) with OpenMP enabled.

### DIFF
--- a/src/PlmBar.f95
+++ b/src/PlmBar.f95
@@ -120,6 +120,7 @@ subroutine PlmBar(p, lmax, z, csphase, cnorm)
             
     scalef = 1.0d-280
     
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Thanks for the input on adding openMP features! Could you please work on the develop branch? I have already implemented several openMP capabilities there and will be merging them to master in the next week or two.

PR #13, opened 2016-05-19
Add omp critical section to PlmBar

opoplawski:


opoplawski:
Moved #12 to develop branch.   Glad to see more openmp work in progress!

MarkWieczorek:
I am curious: Did you try using the develop branch before adding the `omp critical` statements? In comparison to the master branch, in develop, this routine declares a few variables as threadprivate. In particular, on line 73:

~~~~
!$OMP    threadprivate(f1, f2, sqr, lmax_old)
~~~~
I thought that this was all that was needed to make the routine threadsafe, but I could be wrong. If you still really need the `omp critical`, could you perhaps let me know exactly why this is needed? I am a bit busy right now, and if I accept this change, I will make similar changes to all the other routines as well.

MarkWieczorek:
I am going to close this issue. The develop branch will be pushed to master in the coming days. At that point, if OpenMP is not working as it should, please re-open....

opoplawski:
Ah, I hadn't noticed that before.  Yeah, the threadprivate should be sufficient - just alternate ways of achieving the same goal - the options are increased memory usage and initial time computing the private variables vs increased mutex checking on each call to the routine.  Assuming the memory usage is relatively small, I think your method makes more sense.

PR #16, opened 2016-06-16
Remove Makefile redundancies

QuLogic:
The generic rules (e.g. `all`, `clean`, etc.) repeat a lot of the code that's in the Fortran/Python-specific rules. But you can just set a dependency and remove all that redundancy.


QuLogic:
The Makefiles are missing some dependencies that make them less than ideal for parallel builds. This could possible be fixed with a non-recursive build.

Also, the `all`/`all2`/`all3` thing is pretty annoying from a dependency standpoint. Is there a reason you are not doing this with the preprocessor?

MarkWieczorek:
Thanks. I'll start looking at this next week. Totally agree about the all/all2/all3 thing. This was a total hack. Maybe using the preprocessor is a way out. I had to fix a few last things after the 3.2 release, so there are now a few conflicts with your commits.

QuLogic:
> Totally agree about the all/all2/all3 thing. This was a total hack. Maybe using the preprocessor is a way out.

See [this branch](https://github.com/QuLogic/SHTOOLS/tree/dedupe).

> I had to fix a few last things after the 3.2 release, so there are now a few conflicts with your commits.

Rebased.

MarkWieczorek:
I just started looking through your changes, and I see two potential problems.

First, as far as I can tell, LAPACK_FLAGS and FFTW3_FLAGS are not passed to src/Makefile (unless I am missing something). This is easily fixed in the main Makefile.

Second, I use OSX and I am worried about this rule
~~~~
%.o: %.f95
	$(F95) -c $(F95FLAGS) $<
%.o: %.F95
	$(F95) -c $(F95FLAGS) $(LAPACK_FLAGS) $(FFTW3_FLAGS) $<
~~~~
In essence, OSX is **not** case sensitive, but **is** case preserving. (I also wonder how windows machines deal with cases. I don't use it, but I know that some run shtools on it...) Perhaps there is a better way to deal with this on OSX using case insensitive rules? (like mangling the file name some other way?)

QuLogic:
> I just started looking through your changes, and I see two potential problems.

Quite possible; it's a work-in-progress. It does work on my machine, but I'm not sure how portable it is.

> First, as far as I can tell, LAPACK_FLAGS and FFTW3_FLAGS are not passed to src/Makefile (unless I am missing something). This is easily fixed in the main Makefile.

It's strange; I'm pretty sure it worked because my build fails when using the alternate options, but the Make docs say it won't pass variables unless they're `export`ed. Anyway, I guess it won't harm anything to make it explicit.

> Second, I use OSX and I am worried about this rule
> 
> %.o: %.f95
>     $(F95) -c $(F95FLAGS) $<
> %.o: %.F95
>     $(F95) -c $(F95FLAGS) $(LAPACK_FLAGS) $(FFTW3_FLAGS) $<
> 
> In essence, OSX is not case sensitive, but is case preserving. (I also wonder how windows machines deal with cases. I don't use it, but I know that some run shtools on it...) Perhaps there is a better way to deal with this on OSX using case insensitive rules? (like mangling the file name some other way?)

The case thing is a compiler convention to enable the preprocessor. I think it should work on OSX because specfem* uses it and I'm fairly certain people have compiled it successfully on OSX. Windows has odd rules; something like case-preserving only if case is mixed. But it's just a convention, and I think all compilers should have a flag to enable the preprocessor that could be used instead.

PR #17, opened 2016-06-17
Python 3 support

QuLogic:
~~This is based on #16, so the Makefile changes are a bit larger than they should be. I can rebased after that is merged.~~

This is rather straightforward, really only the `print` function and a few minor other changes. The `python` targets will build for both 2 and 3 by default (but this can be changed by setting the `PYTHON_VERSION` variable). To build just one or the other, there are `python2` an `python3` versions of the targets as well.


MarkWieczorek:
I just merged your #16. If you could merge this into your branch, that would help me to go through the new changes. Thanks!

MMesch:
@QuLogic  what would be the best way to integrate this Makefile into a setup.py? Any ideas?

QuLogic:
Rebased.

@MMesch Make and `setup.py` kind of do the same thing, so there's likely always going to be some redundancy. It _is_ just Python though; there's nothing stopping you from shelling out to Make to build the Fortran parts...

MarkWieczorek:
One comment about python3 on OSX. I installed this by brew, and the command in the makefile for setting SYSPY3PATH yields:
~~~~
python3 -c 'import sysconfig; print(sysconfig.get_path("platlib"))')
/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages
~~~~
now, that long path, deep in the cellar of brew, is actually a link to 
~~~~
/usr/local/lib/python3.5/site-packages
~~~~
This should work, but the path is a bit ugly. Not sure if there is anything that should be done about this or not.

MarkWieczorek:
...and if we define SYSPY3PATH this way, why not do the same for SYSPYPATH? (This yields a similar crazy path on OSX).

MarkWieczorek:
Also, in setting the PY3EXT, I get
~~~~
python3 -c 'import sysconfig; print(sysconfig.get_config_var("SO"))'
.cpython-35m-darwin.so
~~~~
Is that really what we want? 

Sorry if these are dumb questions, but I haven't used python3 before.

MarkWieczorek:
I have discovered a small problem with f2py for python3: https://github.com/numpy/numpy/issues/7769

In essence, when I install python3.5, and then use pip3 to install numpy, my pre-existing `f2py` executible gets overwritten by the version 3.5 executible. For whatever reason, they have the same name, whereas it would have been more logical to have used `f2py3`.

So, there is a way around this, and I am curious if this is the way we should do this: we can replace the definitions for F2PY and F2PY3 in the makefile with the following:
~~~~
python -m numpy.f2py
pyhton3 -m numpy.f2py
~~~~
Using these definitions work (including all the test function :) )

Also, there is another odd thing with python3. On my system, f2py3 creates the following folders in the shtools directory:
~~~~
_SHTOOLS.cpython-35m-darwin.so.dSYM/
_constant.cpython-35m-darwin.so.dSYM/
~~~~
It is not clear to me if these are temporary files that should be deleted, or if there is a reason to keep them....

QuLogic:
Sure, it's probably better to define `SYSPYPATH` the same way as well.

The library extension is correct; it is defined by [PEP 3149](https://www.python.org/dev/peps/pep-3149/) to allow for parallel installation of C extensions (and bytecode) for different runtimes.

> we can replace the definitions for F2PY and F2PY3 in the makefile with the following:
> ```
> python -m numpy.f2py
> pyhton3 -m numpy.f2py
> ```

Unfortunately, this is not portable as it depends on which release of NumPy you are using. With 1.9.2, I get:
```
python -m numpy.f2py --quiet -Imodules -Llib --f90flags="-m64 -fPIC -O3 -ffast-math" \
	-c src/pyshtools.pyf src/PythonWrapper.f95\
	-lSHTOOLS -L/usr/local/lib -lfftw3 -lm -llapack  -lblas
/usr/bin/python: No module named numpy.f2py.__main__; 'numpy.f2py' is a package and cannot be directly executed
Makefile:280: recipe for target 'pyshtools/_SHTOOLS.so' failed
make: *** [pyshtools/_SHTOOLS.so] Error 1
```

> It is not clear to me if these are temporary files that should be deleted, or if there is a reason to keep them....

I don't really know. The extension suggests it's macOS-specific. I guess you can see if any other `.so` have a corresponding `.so.dSYM` directory in your existing Python installation?

MarkWieczorek:
Ok. I'm going to pull this as is. Could you do two things?

* Add the directories `_SHTOOLS.cpython-35m-darwin.so.dSYM/z=` and_`constant.cpython-35m-darwin.so.dSYM/` to the `make clean` in one of your later pull requests? These are debug files that were generated by `f2py` on OSX. On other systems, I think that you need to explicitly specify `--debug`, but on my system, these are generated by default.
* Rebase your other pull requests to make my life easier....

My temporary workaround for the OSX numpy bug is 
~~~~
make F2PY="python -m numpy.f2py" F2PY3="python3 -m numpy.f2py"
~~~~

QuLogic:
OK, I'll add that to the dedupe PR when it's ready. The other PRs were rebased.

PR #19, opened 2016-06-17
Dupe consistency

QuLogic:
The duplicate files for different underscore conventions were slightly out of sync with each other.

I still think it is best to get rid of these duplicate files entirely, but I'll leave that for after these PRs are accepted.


QuLogic:
There were also a couple of missing `external`, and the Makefile listed a file in the wrong place.

PR #21, opened 2016-06-25
Use preprocessor to implement underscore name-mangling

QuLogic:
Remove the duplicate files and use the preprocessor to take care of adding an underscore to subroutine names.

This currently uses the filename to trigger the preprocessor, but if that doesn't work, it should be possible to do so using a compiler option.


MarkWieczorek:
This works great on OSX. 

I also tested this using Absoft ProFortran. Absoft gives warnings about quotes that follow comments, such as like
~~~~
f95 -c -m64 -O3 -YEXT_NAMES=LCS -YEXT_SFX=_ -fpic -speed_math=10 -DLAPACK_UNDERSCORE -DFFTW3_UNDERSCORE MakeGravGridDH.F95
MakeGravGridDH.F95:45:76: warning: missing terminating ' character [-Winvalid-pp-token]
!           GM          Product of the gravitatonal constant and the planet's 
                                                                           ^
~~~~
but otherwise it works. I contact their support asking them to fix this.

It remains to be seen how windows machines deal with the `f95` / `F95` naming and preprocessor.

PR #24, opened 2016-07-20
Simple setup.py script

ioshchepkov:
I created a simple setup.py script which can be used instead of 

`make install-python2` or  `make install-python3`

if you want to install only pyshtools without Fortran library.
It works with both python 2 and 3, supports `pip` command.

The main issues:
1. It works only after usual compilation with `make fortran && make python` commands. 
2.  The script contains `get_version()` function for unique (release number from VERSION file + 0.1 + .dev0 + git revision) versioning through development stage with PEP440 support. Other convention, of course, can be made.
3. It's recommended to put `author_email` key in `setup()`

The next stage is to include f2py compilation routines in setup.py. It is not so difficult, except putting documentation from `make_docs.py` to the final package. I really got stuck and maybe will work on it in the future, if no one else will do it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,71 @@
+
+import os
+import re
+import glob
+from setuptools import setup, find_packages
+from subprocess import CalledProcessError, check_output
+
+def get_version():
+    """Get version from git and VERSION file
+
+    Derived from: https://github.com/Changaco/version.py
+    """
+    d = os.path.dirname(__file__)
+    # get release number from VERSION
+    with open(os.path.join(d, 'VERSION')) as f:
+        vre = re.compile('.Version: (.+)$', re.M)
+        version = vre.search(f.read()).group(1)
+
+    if os.path.isdir(os.path.join(d, '.git')):
+        # Get the version using "git describe".
+        cmd = 'git describe --tags'
+        try:
+            git_version = check_output(cmd.split()).decode().strip()[1:]
+        except CalledProcessError:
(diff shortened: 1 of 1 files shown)


MMesch:
Nice ! Two things should be included immediately:

* a simple call to the makefile for python setup.py build should be included. Very simple solution that can be adapted (http://stackoverflow.com/questions/18436061/python-setup-py-call-makefile-dont-include-binaries):
```
directory = 'bin'
if not os.path.exists(directory):
    os.makedirs(directory)

subprocess.call(['make', '-C', 'src'])
```

* the next step is to move the python compilation to the setup.py, which might make things simpler. The reference for f2py is here http://docs.scipy.org/doc/numpy-dev/f2py/distutils.html .

If you prefer, we can also merge first and than add these things later !?

ioshchepkov:
Well, as I said before, I've already done it and have source code. But I have some problems with moving final python docs to the build directory. This I can not simply solve right now. So, yes, I suggest merge this now and then add new things in the future.

MMesch:
OK sounds good! Feel free to push immediately to another PR so that we can follow / help .

MMesch:
One more comment: it seems a MANIFEST file helps setup.py to find non-python files. Did you try this approach?

ioshchepkov:
Yes, and it didn't work for the first try. Actually, my solution worked even with `python setup.py build`, but not with `pip`. It seems like `pip` do not "track" doc files.

PR #37, opened 2016-07-31
Remove future imports from most pyshtools files.

MarkWieczorek:
I've removed most future imports that are not necessary. The only exceptions are a few import print_function when print is explicitly used.

One problem I would like to fix is the import _SHTOOLS statement in make_docs, as this is not a relative import. I tried the following but it doesn't work:

In the make file, I ran this as a module using "-m"
~~~~
$(PYTHON) -m ./pyshtools/make_docs.py . .
~~~~
and in make_docs.py I used
~~~~
from . import _SHTOOLS
~~~~
but then I get the error when building:
~~~~
python -m ./pyshtools/make_docs.py . . 
/usr/local/opt/python/bin/python2.7: Relative module names not supported
~~~~

### DIFF
--- a/pyshtools/__init__.py
+++ b/pyshtools/__init__.py
@@ -34,6 +34,8 @@ and the GitHub project page at
    https://github.com/SHTOOLS/SHTOOLS
 """
 
+from __future__ import absolute_import as _absolute_import
+from __future__ import division as _division
 from __future__ import print_function as _print_function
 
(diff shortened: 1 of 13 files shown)


QuLogic:
You should leave them in for consistency in the future between 2 and 3. I don't see any advantage to only importing them where necessary.

PR #39, opened 2016-08-05
notebook and class interface fixes / coupling matrix and 3d plotting

MMesch:
The 6 notebooks test most of the class interfaces. In this PR, the regional / global spectral comparisons seem to work well, as do the coupling matrix routines / plotting, etc ...


MMesch:
I added a simple 3d plot function:

3dplot

It is explained in the associated notebook...

MarkWieczorek:
So, if i understand, the Fortran file by default now returns the "full" coupling matrix now. I could easily add an optional fortran argument to allow for either "full" or "valid". This would require some conditional tests to determine what the dimensions of the output matrix should be. I am also ok leaving this as is, as long as the rational for this clearly stated in the docs (I'll try to modify this tomorrow or monday).

However, I'm not sure if we can use "if" statements in the python signature file to set the dimensions of the matrix. Thus, we might need to return the full version anyways. What do you think?

MMesch:
I agree. If we can implement it easily we can put it into the Fortran Code
as well. It is only worth doing so if someone Deals with very high degrees
though...
Am 06.08.2016 7:00 nachm. schrieb "Mark Wieczorek" <notifications@github.com

MarkWieczorek:
Could you let me know what the acronym "nl" means here?
~~~~
def __init__(self, tapers, eigenvalues, weights):
        self.kind = 'mask'
        self.nl = _np.sqrt(tapers.shape[0]).astype(int)
~~~~
I'm not quite sure how to describe its use in the documentation. Personally, I think that we can just get rid of it as it is mostly used just to define lmax. If there is a reason to keep it, no problems.

MMesch:
No real reason. NL is the total number of degrees, i.e. lmax + 1 . I found
nl more comfortable to use (lmax+1 appears all the time) and consistent
with the definitions of range than lmax. We can get rid of it if you want.
It might be easier for the documentation.
Am 06.08.2016 21:59 schrieb "Mark Wieczorek" <notifications@github.com>:

PR #41, opened 2016-08-06
cleaner installation instructions

MMesch:
installation instructions are cleaner now. On OSX, installation should be done with ```pip install .``` as well, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,38 +29,32 @@ in spherical harmonics, and multitaper spectral analyses on the sphere.
     OpenMP compatible and OpenMP thread-safe.
 
 ### INSTALLATION ###
-#### OSX ####
-
-```
-brew tap shtools/shtools
-brew install shtools
-```
-
-#### Linux ####
-First, install the dependencies:
+#### Requirements ####
+Linux:
 ```bash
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Concerning the osx install, I would still keep the brew commands, as this is the standard osx package manager (and I'll be maintaining it). The pip install works as on linux, but you need to use brew to install fftw. (I believe that lapack and blas are installed in the base osx system.) You can find my full instructions for this on the updated web documentation:

https://github.com/SHTOOLS/SHTOOLS/blob/develop/www/install.html

MMesch:
Makes sense but keep in mind that apt-get and brew are somewhat outdated to
install Python stuff. Anaconda as first package Manager and pip as second
seems to be the standard and make life a lot easier. At least that was my
impression.
Am 06.08.2016 7:03 nachm. schrieb "Mark Wieczorek" <notifications@github.com

MarkWieczorek:
I'll merge this and then add the brew commands after. The order I'll be using for install priorities will by pip (for python), brew (osx) for fortran+python, make for fortran+python.

PR #43, opened 2016-08-10
Fix some conversion warnings

QuLogic:
This mostly makes explicit things that are already done.
 * REAL->INTEGER are cast with `int`
 * INTEGER->INTEGER*1 are fixed by changing `phase` into a regular INTEGER. This is consistent with some other cases, though I don't know if it should be the other way. It would change the ABI though.

There are three lines with a COMPLEX->REAL conversion that are left, but I'm not sure what the correct fix is there.

### DIFF
--- a/src/MakeCircleCoord.f95
+++ b/src/MakeCircleCoord.f95
@@ -58,7 +58,7 @@ subroutine MakeCircleCoord(coord, lat, lon, theta0, cinterval, cnum)
         interval = 1.0d0
     end if
     
-    num = 360.0d0/interval
+    num = int(360.0d0/interval)
     
     if (present(cnum)) then
         cnum = num
--- a/src/MakeEllipseCoord.f95
+++ b/src/MakeEllipseCoord.f95
@@ -46,7 +46,7 @@ subroutine MakeEllipseCoord(coord, lat, lon, dec, A_theta, B_theta, &
         interval = 1.0d0
     end if
     
-    num = 360.0d0/interval
+    num = int(360.0d0/interval)
     
     if (present(cnum)) then
         cnum = num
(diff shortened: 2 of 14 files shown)


MarkWieczorek:
Looks good to me. My rational for using int*1 is that I thought the multiplications with reals would be faster. Timing test show that this doesn't make any difference with gfortran. 

As for the complex warning:
~~~~
gfortran:f90: src/SHMTVarOpt.F95
    src/SHMTVarOpt.F95:168:35:

                             Fij(i,i) = Fij(i,i) + 2.0d0 * temp1*dconjg(temp1)
                                       1
    Warning: Possible change of value in conversion from COMPLEX(8) to REAL(8) at (1) [-Wconversion]
~~~~
I need to look into this. Obviously, I am just taking the magnitude of a complex number, so the imaginary part will be zero. To get rid of the warning, we could just type cast this as DBLE(). However, we could also use a function like ABS or CABS instead to do this. I've had some bad experiences with compilers screwing up these conversions, as a double precision complex number is 16 bytes, but a double real is 8. The fact that the warning is treating this as a complex(8) bothers me, as the real portion might only by 4 bytes...

MarkWieczorek:
I'll just type cast to the output as dble.  Its possible the warning refers to multiplying 2 * temp1.

QuLogic:
Oh I didn't read that too closely; `temp1*dconjg(temp1)` is of course still complex even if the imaginary part is zero, so assigning to `Fij` (which is a `real*4`) should cause a warning. Using `Abs(temp1)**2` would probably work.

QuLogic:
Oh, also using `dble` that you already did works too.

PR #47, opened 2016-08-17
Pypi fixes

MMesch:
This PR makes SHTOOLS pypi ready. It works on the testpypi server and can be installed with:

``` pip install -i https://testpypi.python.org/pypi pyshtools```

The version string needs to be improved because the direct git version fails for the pypi upload ...

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,7 +9,6 @@ graft examples/ExampleDataFiles
 recursive-include examples/fortran *.f95 input* Makefile
 recursive-include examples/python *.py Makefile
 recursive-include examples/notebooks *.ipynb
-include examples/python/ClassInterface/topo.dat.gz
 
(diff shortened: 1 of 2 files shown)


MMesch:
this shouldn't get merged immediately because it doesn't use the github version string

MMesch:
To repeat the stuff from gitter: we just need to cleanly tag the final version, such that we can use the get_version() function. The download link needs to be adjusted as well, than it should work immediately on pypi.

MarkWieczorek:
Do you want me to merge this with develop? The only issue I see is that this edit needs to be reverted to the url for master
~~~~
download_url='https://github.com/MMesch/SHTOOLS/zipball/pypi-fixes',
~~~~

MMesch:
It can be merged now. I changed the dl-link

PR #49, opened 2016-08-18
phaseonly keyword

MMesch:
This keyword for the ```from_random``` constructor allows to generate coefficients that have random phase but fixed amplitude:

```
clm = pyshtools.SHCoeffs.from_random(power, phaseonly=True)
```

### DIFF
--- a/examples/python/ClassInterface/exact_power.py
+++ b/examples/python/ClassInterface/exact_power.py
@@ -13,12 +13,12 @@ def example():
     degrees[0] = np.inf
     power = degrees**(-1)
 
-    clm1 = pyshtools.SHCoeffs.from_random(power, phaseonly=False)
-    clm2 = pyshtools.SHCoeffs.from_random(power, phaseonly=True)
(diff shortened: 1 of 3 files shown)


MarkWieczorek:
It looks like this change simply normalizes the `random` power at each degree to be exactly the input value. Though the output power should be exactly the same as the input, I think that the term `random phases` means something slightly different.

For me, random phase means that the power is known for each `l` and `|m|`, and the repartition of power between `m` and `-m` is 

````
clm = P * cos(random angle)

slm = P * sin(random angle)
````

MMesch:
If you imagine a set of coefficients as a vector, I think about power as its square length and phase as its direction. The elements of the vector can be for example be the set of all coefficients with a given degree l, or just the two (Re/Im) coefficients with a given l and m. It is a more general perspective of phase. As power, phase can refer to many things.

I agree with you to change it because it is not absolutely clear what it means, and we already use phase in the keyword csphase . What do you think e.g. about ```fixed_power``` ?

PR #51, opened 2016-08-20
WIP

MarkWieczorek:
This pull request is about ready to be merged. There are still a few things I would like to do (see below), and if the consensus is to get a new release out as soon as possible with PyPI support, these could be put off to a future release near the end of the year (i.e., version 3.4, which will fix the Fortran STOP issue).

## Summary of changes:

* Add Earth topography coefficients referenced to mean sea level to the example files, expanded to degree 300: srtmp300.msl. (**Note that the lower resolution topo gridded data in this directory are somewhat redundant. Also, the gridded data in the Examples folder starts at 180 degrees longitude, instead of 0 as used in DH Grids!**)

* Add new function `SHBiasKMask` which is the arbitrary window counterpart to the spherical cap window `SHBiasK`.

* Add `get_biasedpowerspectrum()` method to `SHWindows` for arbitrary windows.

* Add `copy()` method to all classes, which returns a deep copy of the instance.

* Add `__sub__`, `__add__` and `__mul__` operators for two sets of coefficient classes.

* Clean up existing operators for grids, and add the operators `__div__` and `__truediv__`

* Add `__pow__`, `__radd__`, and `__rsub__` operator, and allow most operators to also work with a scalar.

* Add `nwinrot` option when rotating spherical cap windows in `SHWindow` that will rotate only the first `nwinrot` windows.

* Add `degrees` option to `get_lats()` and `get_lons()` methods.

* Add the constructor `from_file()` to initialize an SHGrid with a numpy formatted data file.

* Update Intro 1 notebook and add example to Intro notebook 2 showing how to use arbitrary localization windows.

* Add `tofile()` methods to output raw grid and coefficient data as either text or binary formatted files.

* Add option to read coeffs and grids from a numpy formatted binary file.

* Convert notebooks to html and add links to web documentation.

## To do for future versions ##

* need to think about whether certain methods should return copies of the data, or just references.

* modify `MakeGravGridDH` to also return the potential on the ellipsoid like in Fortran routines. This would have to be a non-optional argument (as opposed to in the Fortran 95 code where it is optional output), and this would hence require calculating 4 grids instead of just 3, which will slow down the code bit.

* Some of the notebooks are redundant. Need to think about re-organizing these into a few simple examples, and a few more in depth tutorials.


README.md
```diff
@@ -77,8 +79,8 @@ well as simple interfaces.
 
 To get started, check out the following Python tutorial notebooks:
 
-* [Introduction 1: A simple introduction to pyshtools: Grids and Coeffs](examples/notebooks/Introduction-1.ipynb)
-* [Introduction 2: A simple introduction to pyshtools: Localization windows and spectral analysis](examples/notebooks/Introduction-2.ipynb)
+* [Introduction 1: Grids and Spherical Harmonic Coefficients](examples/notebooks/Introduction-1.ipynb)
+* [Introduction 2: Localization Windows and Spectral Analysis](examples/notebooks/Introduction-2.ipynb)
 * [tutorial 1: Simple Spherical Harmonic Expansions](examples/notebooks/tutorial_1.ipynb)
```
QuLogic: If Introduction is capitalized, then maybe Tutorial should be too.

pyshtools/shclasses.py
```diff
@@ -288,50 +291,53 @@ def from_random(self, power, kind='real', normalization='4pi', csphase=1):
...
 
         Usage
         -----
 
-        x = SHCoeffs.from_file(filename, lmax, [format, kind, normalization,
-                                                csphase, skip])
+        x = SHCoeffs.from_file(filename, lmax, [format='shtools', kind,
+                                                normalization, csphase, skip])
+        x = SHCoeffs.from_file(filename, [format='npy', kind, normalization,
+                                          csphase, skip])
```
QuLogic: Based on the parameter description below, `skip` should not be necessary with `format='npy'`.

pyshtools/shclasses.py
```diff
@@ -350,21 +356,97 @@ def from_file(self, fname, lmax, format='shtools', kind='real',
...
+
+        Parameters
+        ----------
+
+        filename : Name of the output file.
+        format   : 'shtools' (default) or 'npy'.
+        **kwargs : Keyword arguments of numpy.save().
+        """
+        if format is 'shtools':
+            file = open(filename, mode='w')
```
QuLogic: Better to use a `with` statement.

MarkWieczorek:
I think that this is ready to merge.

PR #58, opened 2016-08-31
Modify travis to run notebooks

MarkWieczorek:
The following modifications have been made:
* The python notebooks are now run by a single python script `test_notebooks.py`, which reads the notebooks, and executes them. This file determines which version of python is being used, and executes the notebooks using that kernel. This is in contrast to the current situation that uses `make notebooks` and which tries to convert all notebooks to html.
* All python tests in `make python-tests-no-timing` are run. This does not run the timing/accuracy routines, which are computationally expensive.
The python tests are run using `Agg` by setting `export MPLBACKEND=Agg` in the travis script.
* In order to avoid installing `tex` on travis, I have commented out the line `'text.usetex': True` in the file `FigStyle.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,14 +53,16 @@ install:
 
 script:
   - if [[ "${PYTHON_VERSION:0:1}" == '2' ]]; then
+      make remove-notebooks
       make notebooks2;
     else
+      make remove-notebooks
       make notebooks3;
     fi
   - mkdir empty
   - cd empty
   - python -c "import pyshtools"
-  - python ../examples/python/TimingAccuracy/TimingAccuracyDH.py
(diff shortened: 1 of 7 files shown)


MarkWieczorek:
Travis encouters this error when running `make python-tests`:

```
RuntimeError: LaTeX was not able to process the following string:
'lp'
```
It is possible that we need to install latex on travis. I am not quite sure how to do this (there is no easy package manager for this on OSX that I am aware of).

MarkWieczorek:
This might be related to 
```
mpl.rcParams.update(style_shtools)
```
in the example files, which defines (in the file `FigStyle.py`)
```
'text.usetex': True,
```
If so, we could either deactivate this parameter, or deactivatee the `rcParams` call in all of the example files.

MMesch:
Let's deactivate this parameter.

PR #59, opened 2016-09-03
Fix Fortran STOP

MarkWieczorek:
This is a working example that fixes Fortran calling STOP, and which then kills the Python kernel. The Python part that raises the error is not too elegant, and I'm looking for suggestions on how to improve this.

In summary:
- The fortran routine (`PlmBar` in this example) has a new optional variable: `exitstatus`. If the conditions for a stop are encountered in the Fortran code, and if the optional argument `exitstatus` is specified, `exitstatus` is returned as a non-zero value.
- The Python wrapper function in `PythonWrapper.f95` calls `PlmBar` with `exitstatus` specified, and checks to see of `exitstatus` is non-zero. If it is, it uses a Python callback function called `pystop` that raises a `RuntimeError`, printing different messages depending on what the value of `exitstatus` is. `pystop` is defined as a simple function in the main `__init__.py` file.

Here is an example of what the output looks like. Note that the first three error lines come from within the Fortran routine, the rest is from Python:

```
In [1]: pysh.shtools.PlmBar(2, -1.2)
 Error --- PlmBar
 ABS(Z) must be less than or equal to 1.
 Input value is   -1.2000000000000000     
capi_return is NULL
Call-back cb_pystop_in_PlmBar__user__routines failed.
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-95c73831c866> in <module>()
----> 1 pysh.shtools.PlmBar(2, -1.2)

/Users/me/SphericalHarmonics/shtools-git/pyshtools/__init__.py in _pystop(status)
    115     elif (status == 2):
    116         raise RuntimeError('Improper bounds for ' +
--> 117                            'input variable.')
    118     elif (status == 3):
    119         raise RuntimeError('Fortran Runtime Error: Error allocating memory.')

RuntimeError: Improper bounds for input variable.
```

I don't like the end result for the user because it provides a traceback that ends in the `pystop` routine, which is not very helpful (that is not where the real error was encountered). If I instead replace the `raise RuntimeError` with a `print` statement, the output from `PlmBar` (which is an array or zeros) will be output, which again, is not useful.

There are different ways that we can handle this exception. Currently, I am using the following in `__init__.py`:

```
def _pystop(status):
    if (status == 1):
        raise RuntimeError('Improper dimensions of input array.')
    elif (status == 2):
        raise RuntimeError('Improper bounds for ' +
                           'input variable.')
    elif (status == 3):
        raise RuntimeError('Fortran Runtime Error: Error allocating memory.')

_SHTOOLS.pystop = _pystop
```

We could instead perhaps define an exception which is raised via like this (see https://github.com/pearu/f2py/wiki/FAQ2ed)

```
class F2PYSTOP(Exception):
    def __call__(self, status):
        raise self.__class__(status)
```

but I haven't been able to get this to work properly (I am slowly learning about exceptions). We could also use `try...except` but I am not quite sure how to implement this within the Fortran `PythonWrapper.f95 file`. We could try to disable the Python traceback, but I haven't been able to figure out how to do this either.

Lastly, a question that needs to be addressed: should we, when possible, check for these errors in the python signature file (pyshtools.pyf)? This would stop execution before entering the Fortran routine, but, in my opinion, the Fortran error message is more informative. In any case, we can not test for all errors in the signature file, so we do need to return `exitstatus`.

Any comments would be welcome!


MMesch:
Looks great! As you said and did, my feeling is also that a custom Exception would work better but that's just a feeling :) Also it is already pretty good as it is I think and could be used like this ...

MarkWieczorek:
Here is another possibility
```
class FortranStop(Exception):
    def __call__(self, status):
        if (status == 1):
            raise self.__class__('Improper dimensions of input array.')
        elif (status == 2):
            raise self.__class__('Improper bounds for input variable.')
        elif (status == 3):
            raise self.__class__('Error allocating memory.')
        else:
            raise self.__class__('Unhandled Fortran 95 error.')

_SHTOOLS.pystop = FortranStop()
```
which results in
```
In [1]: pyshtools.shtools.PlmBar(4, 1.2)
 Error --- PlmBar
 ABS(Z) must be less than or equal to 1.
 Input value is    1.2000000000000000     
capi_return is NULL
Call-back cb_pystop_in_PlmBar__user__routines failed.
---------------------------------------------------------------------------
FortranStop                               Traceback (most recent call last)
<ipython-input-1-ce4e895d3db9> in <module>()
----> 1 pyshtools.shtools.PlmBar(4, 1.2)

/Users/me/SphericalHarmonics/shtools-git/pyshtools/__init__.py in __call__(self, status)
    127             raise self.__class__('Improper dimensions of input array.')
    128         elif (status == 2):
--> 129             raise self.__class__('Improper bounds for input variable.')
    130         elif (status == 3):
    131             raise self.__class__('Error allocating memory.')

FortranStop: Improper bounds for input variable.
```
In this case, the class `FortranStop` will be visible in the main namespace.

Is this better, or worse?

MMesch:
almost no difference. Question of taste in my opinion.

MarkWieczorek:
I could add more checks to the python signature file to avoid calling the fortran routines with invalid parameters that lead to a `stop`. However, I am not sure which of the two error messages is better.

**Fortran error message**
```
In [1]: pysh.shtools.PlmBar(-3,1.1)
 Error --- PlmBar
 LMAX must be greater than or equal to 0.
 Input value is           -3
capi_return is NULL
Call-back cb_pystop_in_PlmBar__user__routines failed.
---------------------------------------------------------------------------
FortranStop                               Traceback (most recent call last)
<ipython-input-1-ae1daeb5c0c8> in <module>()
----> 1 pysh.shtools.PlmBar(-3,1.1)

/Users/me/SphericalHarmonics/shtools-git/pyshtools/__init__.py in __call__(self, status)
    122             errmsg = 'Unhandled Fortran 95 error.'
    123 
--> 124         raise self.__class__(errmsg)
    125 
    126 _SHTOOLS.pystop = FortranStop()

FortranStop: Improper bounds for input variable.
```

**Python signature file error from `check` statement:**
```
In [1]: pysh.shtools.PlmBar(-3,1.1)
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-1-ae1daeb5c0c8> in <module>()
----> 1 pysh.shtools.PlmBar(-3,1.1)

error: (lmax>=0) failed for hidden p_d0: PlmBar:p_d0=1

In [2]: pysh.shtools.PlmBar(-3,1.1)
```
Which of the two is preferable?

andreww:
Another way to do this is to return `exitstatus` back to python from the Fortran and convert the status to an exception there. It is possible to do this with something that looks very much like a decorator (but without using the `@...` syntax). This gives:

```
In [2]: pyshtools.shtools.PlmBar(4, 1.2)
 Error --- PlmBar
 ABS(Z) must be less than or equal to 1.
 Input value is    1.2000000000000000     
---------------------------------------------------------------------------
SHToolsError                              Traceback (most recent call last)
<ipython-input-2-ce4e895d3db9> in <module>()
----> 1 pyshtools.shtools.PlmBar(4, 1.2)

/Users/earawa/Code/test_SHTOOLS/SHTOOLS/pyshtools/__init__.py in wrapped_func(*args, **kwargs)
    129         returned_values = func(*args, **kwargs)
    130         if returned_values[0] != 0:
--> 131             raise SHToolsError(get_status_message(returned_values[0]))
    132         return returned_values[1:]
    133     wrapped_func.__doc__ = func.__doc__

SHToolsError: Improper bounds for input variable.
```

The problem here is that the function that does the wrapping turns up at the inside of the traceback, which isn't ideal. However, I think it should be possible to modify the traceback to point at the next frame (outwards) in which case the error would point to the `pyshtools.shtools.PlmBar(4, 1.2)` line with the same "Improper bounds" error message.

Anyway, the code for the above is almost all in `__init__.py`:

```
class SHToolsError(Exception):
    pass


def get_status_message(status):
    if (status == 1):
        message = 'Improper dimensions of input array.'
    elif (status == 2):
        message = 'Improper bounds for input variable.'
    elif (status == 3):
        message = 'Error allocating memory.'
    else:
        message = 'Unhandled Fortran 95 error.'
    return message


def raise_errors(func):
    def wrapped_func(*args, **kwargs):
        returned_values = func(*args, **kwargs)
        if returned_values[0] != 0:
            raise SHToolsError(get_status_message(returned_values[0]))
        return returned_values[1:]
    wrapped_func.__doc__ = func.__doc__
    return wrapped_func

_SHTOOLS.PlmBar = raise_errors(_SHTOOLS.PlmBar)
```

and the last line would, if this were implemented for all functions, be inside a loop over callable objects in `_SHTOOLS`, much like the approach taken for the documentation. I ended up returning the exit status as the first returned value from Fortran. This looks a bit odd, but I think it probably makes stripping it out easer if we have optional output arguments.

It ought to be possible to make the traceback nicer using the `traceback` module but I have only managed to strip out the outer (not the inner) frame, and the way exceptions work in detail seems to be different between python2 and python3 (which has `.with_traceback()` when raising exceptions). Can I just iterate over the traceback object?

I'm not sure if this is a better solution, but it may be worth thinking about. One final issue: it seems we cannot use `functools.wraps` to keep the function name etc. as the fortran objects don't have all the necessary attributes. We could set what we actually need inside raise_errors, like I currently do for `__doc__`.

andreww:
For the record the change for doing this in Python is now at https://github.com/andreww/SHTOOLS/commit/755a72d4c86620ec5c5690f7c9944ab59ae65b3f. The other advantage of doing it this way is that we don't see the message:

    capi_return is NULL
    Call-back cb_pystop_in_PlmBar__user__routines failed.

but I'm not convinced that this is a big issue.

MarkWieczorek:
I'm debating which to choose. With what you are proposing, the python signature file will need to be modified for all routines in order to return the exit status. This isn't a big deal (just a minor annoyance) given that I need to modify all the wrapper functions anyways. This wouldn't work with the few `shtools` fortran functions that exist, but neither would my initial approach either (I would need to convert these to subroutines).

As for the issue with 
```
capi_return is NULL
Call-back cb_pystop_in_PlmBar__user__routines failed.
```
It is not clear if this is a `f2py` bug or not: see https://stackoverflow.com/questions/31970548/f2py-and-capi-return-is-null-woes

It seems that this *should* be able to be fixed by explicitly defining the callback function in the signature file `python module test__user__routines `, but I can't get this to work (perhaps I gave up to quickly).

For the simple avoidance of the STOP, I think that both approaches should work equally well, and in the end, both raise an exception. I'm looking for a clear advantage of one over the other, but I'm not seeing it...

QuLogic:
Does raising in a callback properly propagate through the external call? We've found that it stops before returning into the external code and just prints things out. However, that was not using f2py but pure C which may handle it correctly.

The decorator-esque version may have a slight advantage in new lines of code. Both options need the addition of the exit status, but since exception handling is in Python, it can be applied a bit more easily:
```python
for func in dir(_SHTOOLS):
    if func.startswith('_'):
        continue
    setattr(_SHTOOLS, func, raise_errors(getattr(_SHTOOLS, func)))
```
(though it may need a slight modification to ensure only the functions are modified if anything else is in `_SHTOOLS`.)

QuLogic:
`raise_errors` should maybe also copy some other `__*name__` properties as well.

MarkWieczorek:
[Sorry for the crazy rebase/merge that screwed up the commits above. It is only the last 4 commits that are related to Fortran stop].

OK. So I fixed the fortran stop issue with a single routine `PlmBar`. The strategy was using my approach where the signature file looks like this
``` 
   #warning "Using deprecated NumPy API, disable it by " \
     ^
    src/_SHTOOLSmodule.c:722:13: error: redefinition of 'pystop_'
    extern void F_FUNC(pystop,PYSTOP) (int *exitstatus_cb_capi) {
                ^
    src/_SHTOOLSmodule.c:108:21: note: expanded from macro 'F_FUNC'
    #define F_FUNC(f,F) f##_
                        ^
    <scratch space>:89:1: note: expanded from here
    pystop_
    ^
    src/_SHTOOLSmodule.c:624:13: note: previous definition is here
    extern void F_FUNC(pystop,PYSTOP) (int *exitstatus_cb_capi) {
                ^
    src/_SHTOOLSmodule.c:108:21: note: expanded from macro 'F_FUNC'
    #define F_FUNC(f,F) f##_
                        ^
    <scratch space>:88:1: note: expanded from here
    pystop_
    ^
    1 warning and 1 error generated.

**Comment by MarkWieczorek** [comment]
**Sorry for the rebase that screwed up the above history**

So, I fixed the fortran stop issue for a single function `PlmBar` using the following approach: The python signature file (`pyshtools.pyf`) looks like this
```
        subroutine PlmBar(p,lmax,z,csphase,cnorm,p_d0)
            fortranname pyplmbar
            real*8 dimension(p_d0),intent(out) :: p
            integer intent(in) :: lmax
            real*8 intent(in) :: z
            integer, optional,intent(in) :: csphase = 1
            integer, optional,intent(in) :: cnorm = 0
            integer, intent(hide),depend(lmax),check(lmax>=0) :: p_d0 = (lmax+1)*(lmax+2)/2
            intent(callback,hide) pystop
            external pystop
            integer exitstatus
            call pystop(exitstatus)
        end subroutine PlmBar
```
the wrapper function looks like this
```
    subroutine pyPlmBar(p,lmax,z,csphase,cnorm,p_d0)
        use shtools, only: PlmBar
        implicit none
        real*8, dimension(p_d0),intent(out) :: p
        integer, intent(in) :: lmax
        real*8, intent(in) :: z
        integer, optional,intent(in) :: csphase
        integer, optional,intent(in) :: cnorm
        integer, intent(in) :: p_d0
        integer :: exitstatus
        external :: pystop
        call PlmBar(p,lmax,z,csphase=csphase,cnorm=cnorm,exitstatus=exitstatus)
        if (exitstatus /= 0) call pystop(exitstatus)
    end subroutine pyPlmBar
```
and the `__init__.py` file contains the line
```
_SHTOOLS.pystop = FortranStop()
```
*However*, when I try to do this to a second function, using the same `pystop` routine, I get some kind of conflict. I don't understand what is going on, but if you remove the comments out the last four lines of the routine `PlmBar_d1` in `pyshtools.pyf`, then you get this error
```
    #warning "Using deprecated NumPy API, disable it by " \
     ^
    src/_SHTOOLSmodule.c:722:13: error: redefinition of 'pystop_'
    extern void F_FUNC(pystop,PYSTOP) (int *exitstatus_cb_capi) {
                ^
    src/_SHTOOLSmodule.c:108:21: note: expanded from macro 'F_FUNC'
    #define F_FUNC(f,F) f##_
                        ^
    <scratch space>:89:1: note: expanded from here
    pystop_
    ^
    src/_SHTOOLSmodule.c:624:13: note: previous definition is here
    extern void F_FUNC(pystop,PYSTOP) (int *exitstatus_cb_capi) {
                ^
    src/_SHTOOLSmodule.c:108:21: note: expanded from macro 'F_FUNC'
    #define F_FUNC(f,F) f##_
                        ^
    <scratch space>:88:1: note: expanded from here
    pystop_
    ^
    1 warning and 1 error generated.
```
I tried defining a `python module __user__routines` interface, but I couldn't get that to work.

Do you have any ideas what is going on? I will look into this more over the next couple of days, but if I don't figure out a solution, i will probably need to implement to method of @andreww

**Comment by QuLogic** [comment]
Can you fix the commits?
```
$ git checkout fix-stop
$ git status
# Make sure you don't have any temporary work you'd like to save (use git stash for that)

$ git reset --hard fe600ce0dc291f28672270d476ef60a9f2e33119
$ git push --force origin fix-stop  # Assuming your fork is named 'origin'
```
You actually did the rebase correctly, but instead of the last step (`git push --force`), you pulled / fetch&merged which brought back all the old commits.

**Comment by MarkWieczorek** [comment]
Numpy issue: https://github.com/numpy/numpy/issues/8288

**Comment by MarkWieczorek** [comment]
I am now using the approach of @andreww, but needed to make a small change. In essence, if the routine returns only a single array, it is returned as an array inside of tuple.
```
In [2]: p = pyshtools.shtools.PlmBar(3,0.5)

In [3]: p
Out[3]: 
(array([ 1.        ,  0.8660254 ,  1.5       , -0.2795085 ,  1.67705098,
         1.45236875, -1.1575162 ,  0.35078038,  1.92130327,  1.35856657]),)
```
and thus, `p[3]` doesn't access the array, but rather the 4th tuple (which does not exist).

I made this minor change to the code that seems to work, but I'm not sure if this is the most elegant solution or not. Any comments?
```
def _raise_errors(func):
    def wrapped_func(*args, **kwargs):
        returned_values = func(*args, **kwargs)
        if returned_values[0] != 0:
            raise SHToolsError(_shtools_status_message(returned_values[0]))
        if len(returned_values) == 1:
            return returned_values[1]
        else:
            return returned_values[1:]
    wrapped_func.__doc__ = func.__doc__
    return wrapped_func
```
Now, both of the following get unpacked correctly
```
p = pyshtools.shtools.PlmBar(3,0.5)
p, dp = pyshtools.shtools.PlmBar_d1(3,0.5)
```

**Comment by andreww** [comment]
Avoiding returning a length one tuple like that seems fine to me and I cannot think of a better way to do this. I wonder if `_shtools_status_message` would be useful as a Fortran (rather than Python) function. It depends on if anybody ever uses the exit status from Fortran I suppose - and this would be something very easy to change later.

**Comment by MarkWieczorek** [comment]
All of the fortran subroutines have been fixed so that when the conditions arise that a stop would be encountered, a non-zero exitstatus is sent to python that then raises an exception. The only remaining problems are that this approach does not work with shtools fortran functions (as they can not return an exitstatus). The following functions might thus still encounter a stop:

```
'MakeGridPoint', 'SHPowerL', 'SHPowerDensityL',
'SHCrossPowerL', 'SHCrossPowerDensityL',
'DownContFilterMA', 'DownContFilterMC', 'SHFindLWin',
'SHSjkPG', 'NormalGravity', 'SHConfidence',
'SHMagPowerL', 'SHPowerLC', 'SHPowerDensityLC',
'SHCrossPowerLC', 'SHCrossPowerDensityLC'
```
Fortunately, these functions are pretty benign, and shouldn't crash under normal conditions. Most of these will also eventually be replaced by python native routines in the near future.

If there are no further issues, I will merge this tomorrow.

MMesch:
My final comment: great! :)

PR #61, opened 2016-09-07
return_coeffs returns copy by default

MMesch:
return_coeffs returns copy by default. First part of #50

### DIFF
--- a/pyshtools/shclasses.py
+++ b/pyshtools/shclasses.py
@@ -1014,7 +1014,7 @@ class SHRealCoeffs(SHCoeffs):
         """Test if class is Real or Complex."""
         return kind == 'real'
 
-    def __init__(self, coeffs, normalization='4pi', csphase=1):
+    def __init__(self, coeffs, normalization='4pi', csphase=1, copy=True):
         """Initialize Real SH Coefficients."""
         lmax = coeffs.shape[1] - 1
         # ---- create mask to filter out m<=l ----
@@ -1023,14 +1023,17 @@ class SHRealCoeffs(SHCoeffs):
         for l in _np.arange(lmax + 1):
             mask[:, l, :l + 1] = True
         mask[1, :, 0] = False
-
         self.mask = mask
         self.lmax = lmax
-        self.coeffs = _np.copy(coeffs)
-        self.coeffs[_np.invert(mask)] = 0.
         self.kind = 'real'
         self.normalization = normalization
         self.csphase = csphase
+
+        if copy:
+            self.coeffs = _np.copy(coeffs)
+            self.coeffs[~mask] = 0.
+        else:
+            self.coeffs = coeffs
 
     def make_complex(self):
         """
@@ -1184,7 +1187,7 @@ class SHComplexCoeffs(SHCoeffs):
         """Check if class has kind 'real' or 'complex'."""
         return kind == 'complex'
 
-    def __init__(self, coeffs, normalization='4pi', csphase=1):
+    def __init__(self, coeffs, normalization='4pi', csphase=1, copy=True):
         """Initialize Complex coefficients."""
         lmax = coeffs.shape[1] - 1
         # ---- create mask to filter out m<=l ----
@@ -1196,11 +1199,15 @@ class SHComplexCoeffs(SHCoeffs):
 
         self.mask = mask
         self.lmax = lmax
-        self.coeffs = _np.copy(coeffs)
-        self.coeffs[_np.invert(mask)] = 0.
         self.kind = 'complex'
         self.normalization = normalization
         self.csphase = csphase
+
+        if copy:
+            self.coeffs = _np.copy(coeffs)
+            self.coeffs[~mask] = 0.
+        else:
+            self.coeffs = coeffs
 
     def make_real(self):
         """
@@ -1941,7 +1948,7 @@ class DHRealGrid(SHGrid):
     def isgrid(grid):
         return grid == 'DH'
 
-    def __init__(self, array):
+    def __init__(self, array, copy=True):
         self.nlat, self.nlon = array.shape
 
         if self.nlat % 2 != 0:
@@ -1960,9 +1967,13 @@ class DHRealGrid(SHGrid):
                              )
 
         self.lmax = int(self.nlat / 2 - 1)
-        self.data = _np.copy(array)
         self.grid = 'DH'
         self.kind = 'real'
+
+        if copy:
+            self.data = _np.copy(array)
+        else:
+            self.data = array
 
     def _get_lats(self):
         """Return the latitudes (in degrees) of the gridded data."""
@@ -2021,7 +2032,7 @@ class DHComplexGrid(SHGrid):
     def isgrid(grid):
         return grid == 'DH'
 
-    def __init__(self, array):
+    def __init__(self, array, copy=True):
         self.nlat, self.nlon = array.shape
 
         if self.nlat % 2 != 0:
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
My only comment concerns the description

```
Called without arguments, this method can be used to return a clean copy of the calling SHCoeff instance.
```

As far as I understand, this will return a new instance but the coeffs attribute will just be a reference to the original coeffs attribute. Though this is a "copy", changing 1 will change the other. There is a method called `copy()`, and this makes a copy in a different memory location so that the two are completely decoupled.

I think that we need to be clear about what we mean by "copy" as this is bound to cause problems, especially for people like me coming from Fortran where assignments are not references.

MMesch:
_get_coeffs returns a true copy. So this one does as well.

https://github.com/MMesch/SHTOOLS/blob/18870fbd4214388a9ad54823b924fe60b2733004/pyshtools/shclasses.py#L1080-L1114

MMesch:
This goes back to the discussion whether references or copys of the arrays should be returned.

MarkWieczorek:
I think that we need to choose a consistent way of making copies and references. As examples, these methods return references of existing data (there might be more....)
* `SHCoeffs.from_array()`
* `SHCoeffs` operators
* `SHGrid.from_array()`
* `SHGrid` operators
* `SHGrid.get_grid()`

If `get_coeffs` returns a copy, for consistency, I think that `get_grid` should as well. 

In my opinion, it might be best to initialize `SHCoeffs.from_array()` and `SHGrid.from_array()`with copies of the input data as well. 

As for the operators, I am not sure what the best is. Should we use copies of all the data in the operatations?

MMesch:
```_get_coeffs``` needs to return a copy because it can change the normalization. Otherwise we will get into a big mess. If the user wants a reference, he should use the attribute ```.coeffs``` directly. This would mean that functions generally return copies and attributes can be used for reference purposes.

**Comment by MMesch** [comment]
as for your other points:

* yes, ```from_array``` should initialise SHGrid and SHCoeff with copies.
* operators except for += -= /= *= should return copies
* functions should return copies

**Comment by MarkWieczorek** [comment]
Could you modify SHGrid so that both `SHGrid.from_array()` and `SHGrid.get_grid()` return copies? As for the operators: it appears that these always return copies as the resultant is placed in a different memory location.

**Comment by MMesch** [comment]
all ```__init__.py``` functions of the subclasses make a copy now. Maybe we should add a parameter to avoid this behaviour because sometimes several copies are internally made now.

**Comment by ioshchepkov** [comment]
1. I have a suggestion. May be it is better to rename `return_coeffs` to `convert` and to add in a doc string the comment above the function ('Convert spherical harmonic coefficients to...')?

2. Why do we really need to copy coeffs in `__init__` ? It is not only time, but memory too. I'm often working with high-resolution models myself and many copies is not good with some machines.

**Comment by MMesch** [comment]
I am thinking that a ```no_copy=True``` keyword would be good. We don't have to copy in the init functions, but the interface is much cleaner if we do. Im ok with calling it ```convert```. It avoids confusion with ```get_coeffs``` as well.

**Comment by MarkWieczorek** [comment]
I am ok with having data copied by default, but to have the option to not copy in certain cases (either internally, or when the user really is concerned about the extra overhead). We just need to add a line to the documentation that describes what the default behavior is. In my opinion, when creating a new class instance, it should not have references to any other data. 

As for `convert`, we used to have this function, but then renamed it, so let me explain why:

`SHCoeffs` and `SHGrid` have the methods `return_coeffs()` and `return_grid()`, and `get_coeffs()` and `get_grid()`. The `return` functions return a class instance, and for the case of `SHCoeffs` you have the option of creating a new class using a different normalization. For the `get` functions, these return only a numpy array (again, with the option of changing the normalization). Simple numpy arrays are necessary when using the Fortran routines, or anything that is not built into the class itself. In my opinion we need both options (working directly with the SHCoeffs.coeffs and SHGrid.data is kind of ugly to me). Nevertheless, I am not entirely happy with the nomenclature `return` and `get`, and am open to suggestions on how these might be renamed.

**Comment by ioshchepkov** [comment]
I think `get` methods can be renamed to `to_array()` by analogy with `from_array()`. And from this method any normalization transformations should be removed.

If we want to get numpy array with the different normalisation, we simply write this
```
coefficients.convert(normalization='ortho').to_array()
```
where `coefficients` is the instance of one of the SHCoeffs subclasses. It also removes duplication. Same for the `SHGrid` class.

In `SHWindow` instead of both `get` and `return` should only stay `return` and methods can be renamed in `get` because there will be no confusion as old `get` methods will be removed. 

Inside all of the classes coefficients and grid data can be called not from method but from the class variable  `self.coeffs` and  `self.grid`.

**Comment by MarkWieczorek** [comment]
I'm for renaming `get_coeffs()` and `get_grid()` in `SHCoeffs`, `SHGrid` and `SHWindow` to `to_array()`. However, I do think that having the option to convert to a new normalization with this method is useful, even if the default it to just give what is in the coeffs attribute.

Given that `SHGrid` does not have a `return_grid` function (this would be redundant with the `copy()` method), I see the logic in renaming `return_coeffs()` to `convert()`. However, `SHWindows` has the methods `return_coeffs` and `return_grid`, both of which returns a new class for taper `i`. Keeping `SHCoeffs.return_coeffs()` would thus be consistent with the `SHWindow.return_coeffs()` method. 

Let me think about this more. I'm willing to be convinced otherwise.

**Comment by MarkWieczorek** [comment]
@MMesch There is a minor PR in your repo with two small changes. 

For simplicity, I suggest that we rename these functions in a different PR. Nevertheless, perhaps we should deal with the redundant copying here. I am ok with a `no_copy` parameter, and I have found the following places that appear (please check!) to be making unnecessary copies: (Line numbers are from `develop` and are probaly off by a few...).

Do you want to do this?

Line 857
```
        coeffs = self.get_coeffs(normalization=normalization.lower(),
                                 csphase=csphase, lmax=lmax)
        return SHCoeffs.from_array(coeffs,
                                   normalization=normalization.lower(),
                                   csphase=csphase)
```                                   
Line 1047:
```
        return SHCoeffs.from_array(complex_coeffs,
                                   normalization=self.normalization,
                                   csphase=self.csphase)
```            
Line 1118
```
                return SHCoeffs.from_array(
                tempcoeffs, normalization=self.normalization,
                csphase=self.csphase)
```
Line 1139
```
        gridout = SHGrid.from_array(data, grid='DH')
```        
Line 1160
```
        gridout = SHGrid.from_array(data, grid='GLQ')
```        
Line 1123
```
        return SHCoeffs.from_array(real_coeffs,
                                   normalization=self.normalization,
                                   csphase=self.csphase)
```                                   
Line 1329
```
        return SHCoeffs.from_array(coeffs_rot,
                                   normalization=self.normalization,
                                   csphase=self.csphase)
```
Line 1348
```
        gridout = SHGrid.from_array(data, grid='DH')
```        
Line 1369
```
        gridout = SHGrid.from_array(data, grid='GLQ')
```        
Line 1982
```
        coeffs = SHCoeffs.from_array(cilm,
                             normalization=normalization.lower(),
                             csphase=csphase)
```
Line 2067
```
        coeffs = SHCoeffs.from_array(cilm,
                                     normalization=normalization.lower(),
                                     csphase=csphase)
```
Line 2157
```
        coeffs = SHCoeffs.from_array(cilm,
                                     normalization=normalization.lower(),
                                     csphase=csphase)
```
Line 2242
``` 
       coeffs = SHCoeffs.from_array(cilm,
                                     normalization=normalization.lower(),
                                     csphase=csphase)
```
Line 2629
```
        return SHCoeffs.from_array(coeffs,
                                   normalization=normalization.lower(),
                                   csphase=csphase)
```                                   
Line 2673
```
            return SHGrid.from_array(self.get_grid(itaper, grid=grid.upper(),
                                                   zeros=zeros),
                                     grid='GLQ')
```

**Comment by MMesch** [comment]
Finally I called ```no_copy``` ```copy```. I wasn't sure if we need it for the windows. Probably not but it is implemented nevertheless now. Tell me when we can merge this @MarkWieczorek .

**Comment by MarkWieczorek** [comment]
Need to add `copy` to docstrings and to revisit the purpose of 
```
if copy:
    self.coeffs = _np.copy(coeffs)
    self.coeffs[~mask] = 0.
else:
    self.coeffs = coeffs
```

PR #65, opened 2016-11-03
Fix powerspectrum

MarkWieczorek:
This pull request makes several modifications to the powerspectrum methods of `SHCoeffs` and `SHWindow`.

The following methods were renamed:
* SHCoeffs.get_powerperdegree -> SHCoeffs.powerspectrum
* SHCoeffs.plot_powerperdegree -> SHCoeffs.plot_powerspectrum
* SHWindow.get_powerperdegree -> SHWindow.powerspectra
* SHWindow.plot_powerperdegree -> SHWindow.plot_powerspectra

These methods now all take the following options:
* `unit` = `per_l`, `per_lm`, or `per_dlogl`
* `base` (default 10) for plotting log-log plots and for calculating the power per log bandwidth `per_dlogl`.
* `energy` (default False) to calculate the energy spectrum instead of power spectrum.

Furthermore, 
* The methods `SHCoeffs.get_powerperband` and `SHCoeffs.plot_powerperband` were removed. 
* The web documentation was updated to give the explicit formulas for the power spectra for all normalizations.
* doc strings of wrapped routines were slightly modified by removing the first two lines. For the constants, the header 'Value' was also removed for clarity.


MMesch:
Shouldn't we name the routine: get_spectrum ? It is more consistent with the ```energy=True``` option in my opinion, and we could integrate the L2 norm in it as well. Also, we could keep the underlying methods ```powerperdegree``` and merely call them from the general method get_spectrum. This keeps everything nicely organized.

MarkWieczorek:
In thinking about the l2 norm more, its not clear to me why someone would use this. Can you give me an example? For the case of 4pi normalized coefficients, the l2 norm *is* `powerspectrum()`. For orthonormalized coefficients, the l2 norm is `powerspectrum(energy=True)` (i.e., it differs by a constant). If you were to use schmidt semi-normalized coefficients, you would probably be incorrect in using the l2 norm in the first place given that the power of the spherical harmonic coefficients has a dependence on degree. If you still think that this is necessary, I would propose adding the option `unit=l2norm` (in addition to `per_l`, `per_lm`, and `per_dlogl`). What do you think?

As for renaming `.powerspectrum()` to `.spectrum()`, this has some logic to it given that we have an optional argument that lets you calculate the "energy spectrum". However, I'm not really sure if people who actually calculate the energy spectrum call it the energy spectrum (they probably call it the powerspectrum, and redefine the powerspectrum to be different than ours by 4pi.) In any case, if we do this, then we will also have to rename all the following routines, and I'm not sure if we can do that in a self consistent way. The biggest problem is that I don't know what we could call `plot_power()`. Lastly, I think that everyone knows intuitively what the power spectrum is, but the "spectrum" is a bit ambiguious. We will also have to define a `crossspectrum`.

## SHCoeffs
* powerspectrum() -> spectrum()?
* plot_powerspectrum() -> plot_spectrum()?
* plot_power() -> ?

## SHWindow
* powerspectra() -? spectra()?
* biasedpowerspectrum() ->biasedspectrum()?
* multitaperpowerspectrum() ->multitaperspectrum()?
* multitapercrosspowerspectrum() ->multitapercrosssspectrum()?
* plot_powerspectra() -> plot_spectra()?

MMesch:
I am for renaming. ```get_powerspectrum(energy=True)``` is not very logical to me if we keep using our strict definition of power. I would add ```l2norm```. We don't know what meaning people associate with it. E.g. Schmidt-Semi coefficient value would correspond to something like the ```power per degree l``` if all coefficients had this value. I am just saying, that some people introduced different normalisations because the L2 norm of the coefficients have specific meanings.

PR #71, opened 2016-11-14
Read ICGEM-format

ioshchepkov:
This is an ICGEM-format parser both for main (2011 year) version and for version 2.0.

The function is somewhat large and more universal than the other input functions. It alone can be used for reading both coefficients as well as their errors. It also computes periodical parts of the coefficients on a given epoch in YYYYMMDD.DD format.

## Checklist

- [x]  remove usecols and add errors attribute, which can be either 'formal', 'calibrated' or None (default)
- [x]  return the coefficients and errors in one call
- [x]  fix docs and typos
- [x] move `other` to `utils`


MarkWieczorek:
My suggestion is to keep the old mixed-case naming system only for python-wrapped fortran routines. All python-wrapped fortran routines are found in the `shtools` subpackage, and they are also duplicated in subpackages such as `shio`. 

Once there is a python routine that replaces the functionality of the fortran one, I suggest that we remove it from the relevant subpackage (e.g., `shio`) but keep the old fortran routine in the `shtools` subpackage.

I think that the first priority is to rewrite all the IO routines in `shio`. I dedicated routine for ICGEM is probably needed, but the others (shread, shreadh, shreaderror, shreaderrorh) could probably be combined into a single, relatively generic routine. Based on what we did with the `SHCoeffs.from_file()` method, we might want to make a generic routine like`shread()`, or `read_coeffs()`, etc. and then have an optional arguement that specicies for the format:
```
format : str, optional, default = 'shtools'
    'shtools' format or binary numpy 'npy' format.
```

ioshchepkov:
I renamed the main function to `read_icgem_gfc` and moved it to the `pyshtools.shio` subpackage in `icgem.py` and `common.py` submodules. 

Also I created subpackage `utils` for common auxiliary routines. For example, now it has `datetime.py` submodule to store all date and time related routines. Hope to see your opinion on this, because I'm not sure.

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,148 @@
...
+            raise ValueError('Epoch must be specified for the "icgem2.0" format.')
+        elif epoch is not None:
+            epoch = _yyyymmdd_to_year_fraction(epoch)
+
+        if 'earth_gravity_constant' in header:
+            gravity_constant = float(header['earth_gravity_constant'])
+        elif 'gravity_constant' in header:
+            gravity_constant = float(header['gravity_constant'])
+        else:
+            raise ValueError('No standart gravitational constant in the\
```
MarkWieczorek: typo - No standard

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,148 @@
...
+    Parameters
+    ----------
+    filename : str
+        The ascii-formatted filename containing the spherical harmonic coefficients.
+    usecols : sequence, optional
+        Which two columns to read. Can be used to read errors of the
+        coefficients instead of the coefficients themself. Can be (3, 4) or
+        (5, 6) or (7, 8). Default is (3, 4), i.e. read coefficients.
+    lmax : int, optional
+        Maximum degree to read from file, if None then maximum degree of the
```
MarkWieczorek: Maybe this is clearer? 

Maximum degree to read from the file. If lmax is None, less than 0, or greater than lmax_model, the maximum degree of the model will be used.

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,148 @@
...
+        Standart gravitational constant of the model, in m**3/s**2
+    r0 : float
+        Reference radius of the model, in meters.
+
+    Parameters
+    ----------
+    filename : str
+        The ascii-formatted filename containing the spherical harmonic coefficients.
+    usecols : sequence, optional
+        Which two columns to read. Can be used to read errors of the
```
MarkWieczorek: It might be a good idea to note what the columns correspond to here.

pyshtools/shio/common.py
```diff
@@ -0,0 +1,28 @@
+"""
```
MarkWieczorek: I wonder if this function should be in utils, or if the single function in utils should be simply in shio. Not sure what the best approach is.

pyshtools/shio/common.py
```diff
@@ -0,0 +1,28 @@
...
+Common routines for I/O
+"""
+
+import numpy as _np
+
+def _derivative(epoch, ref_epoch, trnd, periodic):
+    """Return sum of the time-variable part of the coefficients
+
+    The formula is:
+    G(t)=G(t0) + trnd*(t-t0 ) +
```
MarkWieczorek: G(t) = G(t0) + trnd*(t-t0)

pyshtools/shio/common.py
```diff
@@ -0,0 +1,28 @@
+"""
+Common routines for I/O
+"""
+
+import numpy as _np
+
+def _derivative(epoch, ref_epoch, trnd, periodic):
```
MarkWieczorek: Does this routine calculate a "derivative"? Not sure if I understand the naming...

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,188 @@
...
+import numpy as _np
+
+from pyshtools.utils.datetime import _yyyymmdd_to_year_fraction
+
+
+def _time_variable_part(epoch, ref_epoch, trnd, periodic):
+    """Return sum of the time-variable part of the coefficients
+
+    The formula is:
+    G(t) = G(t0) + trnd*(t-t0 ) +
```
MarkWieczorek: extra space after t0...

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,188 @@
...
+        Reference radius of the model, in meters.
+    errors : array, optional
+        Array with the errors of the coefficients with the shape
+        (2, lmax + 1, lmax + 1) for the given epoch.
+
+    Parameters
+    ----------
+    filename : str
+        The ascii-formatted filename containing the spherical harmonic coefficients.
+    errors : str, optional
```
ioshchepkov: I think the equal sign is misleading here, because what exactly is "default" here from a pythonic point of view? Attribute? Variable?

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,188 @@
...
+                periodic_sum += coeffs * _np.cos(2 * _np.pi / period * delta_t)
+            elif trifunc == 'asin':
+                periodic_sum += coeffs * _np.sin(2 * _np.pi / period * delta_t)
+    return trend + periodic_sum
+
+
+def read_icgem_gfc(filename, errors=None, lmax=None, epoch=None):
+    """Read spherical harmonic coefficients from an ICGEM GFC ascii-formatted file.
+
+    This function only reads files with the gravity field spherical
```
ioshchepkov: It does, the ICGEM-format can also be used for ocean/atmospheric tides, but this function will not be able to read it.

pyshtools/shio/icgem.py
```diff
@@ -0,0 +1,188 @@
...
+
+        lmax_model = int(header['max_degree'])
+        if lmax is None or lmax < 0 or lmax > lmax_model:
+            lmax = lmax_model
+
+        if errors is not None:
+            valid_err = ('calibrated', 'formal', 'calibrated_and_formal')
+            if header['errors'] == 'no':
+                raise ValueError('This model has no errors.')
+            elif errors not in valid_err[:-1]:
```
ioshchepkov: No, 'calibrated_and_formal' is not implemented.

'calibrated', 'formal' and 'calibrated_and_formal' are parameters in the ICGEM formatted file for keyword 'errors' which has nothing to do with the parameter `errors` of the function.

'calibrated' or 'formal' mean there are only two columns with the errors - 5 and 6, which will be read if  `errors="calibrated"` or `errors="formal"` respectively. 

'calibrated_and_formal', in the other hand, means there are four columns in the file: 5, 6, 7 and 8. 5 and 6 stand for calibrated errors and 7 and 8 stand for formal. So, the function will return calibrated errors if `errors="calibrated"` or formal errors if `errors="formal"` as before.

I don't know if returning of both calibrated and formal errors is really needed. If it does, the I will write it.


This part is checking that `errors` has valid values and the given value to this parameter is actually exists in the file.

pyshtools/shio/icgem.py
ioshchepkov: Fixed.

PR #74, opened 2016-12-01
Parallel build and flexible installation directory

heroxbd:
Hi, these commits enable parallel build and flexible installation directory, please review.

PR #75, opened 2016-12-01
Parallel build and flexible installation directory

heroxbd:
Hi, these commits enable parallel build and flexible installation directory, and set PY3EXT to nopy3 if python3 is not found or abnormal.

Please review.

See PR #74.


heroxbd:
In PR #75.

MarkWieczorek:
Looks good to me. But maybe we should add DESTDIR and PREFIX to the list of allowable options in the comments at the head of the file so that people know how to use these.

heroxbd:
Thanks Mark.  Excuse me for not reacting in time.

PR #92, opened 2017-05-11
Fix cross_spectrum without optional argument

andreww:
Without the optional argument "degrees" in cross_spectrum
degrees is set using numpy.range(lmax+1). However, this
is not a numpy function and 'nrange' should be used. This
patch makes cross_spectrum work as documented (i.e. without
the degrees argument) and matches the spectrum function.


andreww:
This probably belongs on master, not develop, as it is a bug fix for 4.0. It would be nice to add a test too.

MarkWieczorek:
Give me about a week to look at this. Kind of busy right now...

MarkWieczorek:
I really apologize for how long it took me to get to this. My desk is mostly clean now, and I am planning on fixing a bunch of other things here in the coming months.

PR #97, opened 2017-09-02
Add space behind fortran stop statements

strawpants:
In the developer version only one file remained affected (which is part of the test suite). Basically gfortran was complaining about missing spaces behind stop statements. Sounds more like a compiler bug to me, nut adding a space behind a stop statement does not do any harm either.

### DIFF
--- a/examples/fortran/SHLocalizedAdmitCorr/LocalizedAdmitCorr.f95
+++ b/examples/fortran/SHLocalizedAdmitCorr/LocalizedAdmitCorr.f95
@@ -43,7 +43,7 @@ Program LocalizedAdmitCorr
 	allocate(pot(2,degmax+1,degmax+1), stat = astat(2))
 	if (astat(1) /= 0 .or. astat(2) /= 0) then
 		print*, "Problem allocating arrays TOPO and POT", astat(1), astat(2)
-		stop
+		stop 
 	endif
 
 	print*, "Reading = ", topography_file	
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Could you let me know why you need to add spaces after these `stop` statements? The reason I ask is that I don't get any warnings with gfortran on macosx, and all of the files in the `src` directory have `stop` statements *without* spaces after them.

Also, it would be great if you could add a couple of lines to the wiki on github describing how to install shtools on archlinux.

strawpants:
This is essentially the error I got by compiling with gfortran 7.1.1:
>>
SHLocalizedAdmitCorr/LocalizedAdmitCorr.f95:113:6:

   stop
      1
Error: Blank required in STOP statement near (1)
SHLocalizedAdmitCorr/LocalizedAdmitCorr.f95:162:6:

   stop
      1
Error: Blank required in STOP statement near (1)
>>

It looks like this isssue is mentioned here as well https://www.mail-archive.com/gcc-bugs@gcc.gnu.org/msg526359.html

I'll add a few lines to wiki on how to install on arch

MarkWieczorek:
So if I understand, you can not compile the example programs if they don't have a blank after the `stop`, but you can compile the SHTOOLS library, of which all the files have `stop` statements without blanks? I'm happy to change this for compatibility, but I don't understand why you are not having problems compiling libSHTOOLS.

strawpants:
I honestly don't understand it either, and can't seem to reproduce it for a simple example. I just know that it eliminated the compile error. Anyway it's up to you to decide to pull it.

MarkWieczorek:
The problem was a tab stop following the `stop` statement. I cleaned up the file and will merge this with my own pull request.

PR #101, opened 2017-10-23
BLD: autoconfigure lapack

xoviat:
This PR uses numpy's autoconfiguration feature in order to correctly detect the lapack version that we are linking against. This is needed so that we can potentially link against OpenBLAS, MKL, or other BLAS options.


xoviat:
Bad idea, I'll just patch setup.py

PR #108, opened 2017-11-29
[setup] autoset numpy version

xoviat:
This PR automatically sets the required numpy version to that which shtools is compiled against.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,7 +173,7 @@ KEYWORDS = ['Spherical Harmonics', 'Spectral Estimation', 'Wigner Symbols',
 
 
 INSTALL_REQUIRES = [
-    'numpy (>=' + str(numpy.__version__) + ')',
+    'numpy>=' + str(numpy.__version__),
 ]
 
 # configure python extension to be compiled with f2py
@@ -279,7 +279,7 @@ metadata = dict(
     author_email="mark.a.wieczorek@gmail.com",
     license='BSD',
     keywords=KEYWORDS,
-    requires=INSTALL_REQUIRES,
+    install_requires=INSTALL_REQUIRES,
     platforms='OS Independent',
     packages=setuptools.find_packages(),
     classifiers=CLASSIFIERS,
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ import sysconfig
 # the setuptools import dummy patches the distutil commands such that
 # python setup.py develop works
 import setuptools  # NOQA
+import numpy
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -277,7 +279,7 @@ def configuration(parent_package='', top_path=None):
     author_email="mark.a.wieczorek@gmail.com",
     license='BSD',
     keywords=KEYWORDS,
-    requires=INSTALL_REQUIRES,
+    install_requires=INSTALL_REQUIRES,
```
QuLogic: That doesn't make sense; you need a runtime requirement after it's built, not an install-time requirement.

xoviat: Runtime requirements are deprecated.

xoviat: There are to phases to Python distribution:
1. The build phase, which can specify requirements with PEP 518.
2. The install phase, which can specify requirements with METADATA.

If we build against a certain numpy version, then we require a version of numpy equal or greater to that version to be installed.

QuLogic: Yes, I understand why this PR is necessary.

Sorry, I was confusing `install_requires` with `setup_requires` here.

xoviat:
OSX failure is unrelated.

MarkWieczorek:
OSX Brew problems:

First off, I have no idea what the OCLint library is, since we never installed it. We could probably delete the offending file, but perhaps it is better to simply force the symlinks. I found this link https://github.com/Homebrew/brew/issues/1742, which is kind of a hack, but might work. Could you try replacing the line

````
brew update; brew install fftw --with-fortran
````
with this?
````
brew update; brew install gcc || brew link --overwrite gcc; brew install fftw --with-fortran
````
Installing fftw forces to install the gcc dependency first, so if we build gcc first and then force symlinks, we shouldn't generate any errors. As noted in the link, this is a non-ideal solution, because installing gcc could give rise to errors unrelated to linking.

Also, I think that fftw was probably built ok, as it was linked to the libraries in the cask, not to /usr/local/lib. The problem is that the brew returned an error (which is not important) that forced travis to quit.

xoviat:
Go ahead and push a commit to develop. You'll probably have better luck on this matter than I will.

MarkWieczorek:
The fix from above works: https://github.com/SHTOOLS/SHTOOLS/pull/109
I could merge this to develop, but maybe it would be easier for you to just add the two lines to the .travis file.

xoviat:
Just merge into develop and then I will restart.

MarkWieczorek:
done...

PR #114, opened 2017-12-23
Complete rewrite of html documentation

MarkWieczorek:
This pull request represents a complete rewrite of the html documentation for shtools.

* All html source files have been converted to markdown.
* The markdown source files are now located in `www-src`, specifically, in the directory `pages/mydoc`.
* A static html web site is built using `jekyll`, whose files are located in `doc`. Github will automatically create the static pages and serve them on [shtools.github.io/SHTOOLS](https://shtools.github.io/SHTOOLS/). To build the static pages yourself, it is only necessary to execute `bundle exec jekyll build` in the directory `doc`, which will build the site into `_site` in the same directory. Alternatively, `make www` in the main directory will create a static site in the top-level directory `www` that could be used to deploy on a different web server.
* The site is based on the template [Jekyll documentation theme](https://github.com/tomjoht/documentation-theme-jekyll) by @tomjoht . I have simply deactivated all the bells and wistles, like posts, commenting via disqus, glossary, tooltips, etc.
* The site has a built in search engine which is useful for searching for a specific routine. This is powered by javascript.
* Much of the text has been updated based on the SHTOOLS manuscript that is 98% finished....

There are still a few things I would like to do:

* Creating of the pdf documentation is imperfect: The top level headers are printed twice, MathJax/Latex equations don't work, and some pages are not properly cross-referenced.
* I would like to clean-up/reformat all of the old documentation md files to be web friendly. Before, they were designed only to make unix man pages and python doc strings. However, this is a lot a work, and I don't know if the investment in time is worth it. 
* Javascript is used to power the collapsing menus and search engine. I don't understand how JS works, and I only note here that macOS Safari doesn't seem to support these features.


QuLogic:
Any rendered example?

MarkWieczorek:
For now, you can look at the file `www/index.html` in the PR. I'll push this to the gh-pages branch in the next day or two after I figure out how github autobuilds jekyll projects.

Issue #121, opened 2018-02-07
encoding error in pyshtools/constant/__init__.py


dogod621:
Hi,

I install pyshtools on windows, python3.6.

And when I call 
```
import pyshtools
```
I got an encoding error:
```
Traceback (most recent call last):
  File "C:\Users\CGlab_Server\Desktop\test\sht.py", line 2, in <module>
    import pyshtools as st
  File "C:\Users\CGlab_Server\AppData\Local\Programs\Python\Python36\lib\site-packages\pyshtools\__init__.py", line 48, in <module>
    from . import constant
  File "C:\Users\CGlab_Server\AppData\Local\Programs\Python\Python36\lib\site-packages\pyshtools\constant\__init__.py", line 68, in <module>
    _pydoc = _pydocfile.read()
UnicodeDecodeError: 'cp950' codec can't decode byte 0xe2 in position 220: illegal multibyte sequence
```

I change line 67 in pyshtools\constant\__init__.py to
```
with open(_path, encoding = 'utf8') as _pydocfile:
            _pydoc = _pydocfile.read()
```

then the error is fixed.

MarkWieczorek:
Could you let me know if `utf-8` works as well on windows? This seems to be the standard notation in the docs.

dogod621:
i don't know what are these. 
So I just print _pydoc out for you ( After add encoding = 'utf8' ).

for example: 
_name: 
```
wgs84_gma 
```
_pydoc:
```
The GM of the atmosphere adopted by the WGS84 model.

3.50e8 m3 s-2

Reference
---------
National Imagery and Mapping Agency (2000). Department of Defense World Geodetic
System 1984: Its Definition and Relationship with Local Geodetic Systems. NIMA
TR8350.2, National Imagery and Mapping Agency.
```

MarkWieczorek:
could you let me know if this works?

````
with open(_path, encoding = 'utf-8') as _pydocfile:
            _pydoc = _pydocfile.read()
````
This seems to be the standard encoding to use.

dogod621:
yes, it works.
I print the file string out in my previous comment.
it seems it works.

MarkWieczorek:
Here is my fix, which will be merged into develop later

https://github.com/SHTOOLS/SHTOOLS/pull/123

Let me know if there are any other problems regarding the doc files.

PR #123, opened 2018-02-11
Force doc files to be read as utf-8

MarkWieczorek:
This PR

* removes some non-standard ascii dashes in the documentation files
* forces all doc files to be opened as `utf-8`
* Fixes a few typos in the documentation.

Non ascii characters were causing pyshtools to fail when importing on some windows machines (https://github.com/SHTOOLS/SHTOOLS/issues/121).

### DIFF
--- a/docs/css/customstyles.css
+++ b/docs/css/customstyles.css
@@ -404,7 +404,7 @@ ul#mysidebar {
 }
 
 .nav li a:hover {
-    background-color: #8D8D8D;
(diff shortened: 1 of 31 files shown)


MarkWieczorek:
It appears that the statement

`with open(_path, encoding = 'utf-8')`

breaks compatibility with Python 2.7. Is this the time to end Python 2.7 support?

PR #150, opened 2018-06-04
Add colorbar to map plot

sstaehler:
add a colorbar to the map plot of shcoeffsgrid

also, there were two references to np, while numpy was imported as hidden _np

### DIFF
--- a/pyshtools/shclasses/shcoeffsgrid.py
+++ b/pyshtools/shclasses/shcoeffsgrid.py
@@ -989,9 +989,11 @@ class SHCoeffs(object):
                 angles = _np.array([alpha, beta, gamma])
         elif convention is 'x':
             if body is True:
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Thanks for looking into this, and fixing the np bugs. Just a couple of comments:

* Could you base this off of `develop` and not `master`?

* The second colorbar call doesn't set the orientation, and is hence plotted oppositely as when ax is specified. This makes me think that in the main `plot` method we should specify optional parameters `colorbar=False` (default) and `orientation=horizontal` (default) that get passed to the `_plot()` methods.

* We will need to modify the plot method for complex data, as well as for GLQ grids. Perhaps it will be possible to modify this only once in the main `plot` method and not the `_plot` methods in the subclasses?

* Is there a way to ensure that the colorbar has the same width as the plot? Perhaps this is too complicated.

MarkWieczorek:
The addition to add colorbars have been added to the gravmag PR (in progress).

PR #159, opened 2018-07-01
Add the option of using colatitude instead of latitude in SHCoeffs.expand()

MarkWieczorek:
One can now specify `colat` instead of `lat` for the method `SHCoeffs.expand()`.


MarkWieczorek:
Included in this PR: https://github.com/SHTOOLS/SHTOOLS/pull/162

PR #163, opened 2018-07-18
 Remove some leftover www install lines from install-fortran target

strawpants:
I've removed 3 (www install) lines from the install-fortran target which fails when no website is explicitly build with Jekyll.
I don't know how you prefer to handle this but I guess one could either:
- not install the static website as additional documentation (as with this pull request)
- Add the static website again in the repo
- create a dependency on Jekyll and add the www target as a dependency for the install-fortran target

Regards,
Roelof


MarkWieczorek:
Thanks for noticing this.  We work on the develop branch, so could you resubmit this based on develop?

Also, a couple of these changes have already been made in this big PR that I am going to merge in a few days: https://github.com/SHTOOLS/SHTOOLS/pull/162. If you could make a quick look there first, that would be great to avoid any merge conflicts.

For info, I am aiming for a new release before sept. 1

strawpants:
I see now that this issue is in fact already coped with in 6a1b645b1f2aebb5f3c36041ab1d9715c41e50b9, so you can ignore this PR. I'll check the develop branch next time.

PR #178, opened 2019-02-22
Add methods to_netcdf() and to_xarray() to the class SHGrid

MarkWieczorek:
This PR adds support for saving gridded data in an `SHGrid` class instance to (1) netcdf and (2) xarray formats.
* `to_netcdf()` - export data to netcdf format. When saved to file, these can be used directly with GMT (generic mapping tools), where they are known as 'grd' files. 
* `to_xarray()` - export data to an xarray DataArray.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ install:
         jupyter
         mkl
         astropy
+        xarray
   - conda install -n test-environment -c eumetsat fftw3
   - source activate test-environment
   - pip install --no-deps .
--- a/docs/pages/mydoc/python-shgrid.md
+++ b/docs/pages/mydoc/python-shgrid.md
@@ -56,7 +56,7 @@ table:nth-of-type(n) th:nth-of-type(2) {
 | `to_array()` | Return a numpy array of the gridded data. |
 | `to_xarray()` | Return the gridded data as an xarray DataArray. |
 | `to_file()` | Save raw gridded data to a text or binary file. |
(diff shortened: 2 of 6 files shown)


leouieda:
@MarkWieczorek it would be great to have a `SHGrid.to_dataset()` method or a property to get access to the `Dataset`. It's really convenient for doing computations.

MarkWieczorek:
I was thinking about that too, but I wasn't sure what to name the function, and whether to output a DataArray or Dataset. I could see:

* to_dataarray() - to output an xarray DataArray
* to_dataset() - to output an xarray Dataset, or
* to_xarray() - perhaps with an optional argument to specify whether you want a dataarray or dataset.

I haven't worked with xarray much, so which is best for you?

leouieda:
The `DataArray` is what is used to represent a single variable in the netCDF but a lot of the higher level functionality is in the `Dataset`. You can think of them as `DataArray -> pandas.Series` and `Dataset -> pandas.DataFrame`. I would suggest doing the `Dataset` version only because the underlying `DataArray` can always be accessed from that. Plus, you could easily concatenate multiple grids by merging the `Dataset`.

Maybe `to_dataset` would be good since there is already `to_array`.

MarkWieczorek:
I've added the method `to_xarray()`. This is the same name convention as used in pandas. Also, I've chosen to output the data as a DataArray, which is most similar to what is the default behavior of pandas.

PR #195, opened 2019-09-24
Remove Python 2.7 compatibility from pyshtools

MarkWieczorek:
From this point onwards, pyshtools will not support python 2.7. This pull request removes all Python 2.7 specific instructions from the code base (such as calls to `__future__`). The Python2.7-compatible code will no longer be actively developed, and only critical updates will be made on the branch `python2.7`.

The pull request also makes a few small changes to the setup.py file.
* The `get_compiler_flags` function was removed as it was never used. The keyword arguement was misspelled, so adding flags to this has no effect on compilation.
* For windows (`win32`) I have set the `runtime_library_dir` to `[]` based on some discussions I've found on other projects. This will hopefully get the appveyor windows builds to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,9 @@ language: c
 
 matrix:
   include:
-    # Test Python 2.7 and 3.7 on Linux and OSX
-    - os: linux
-      env: PYTHON_VERSION=2.7
+    # Test Python 3.7 on Linux and OSX
     - os: linux
       env: PYTHON_VERSION=3.7
-
-    - os: osx
-      env: PYTHON_VERSION=2.7
     - os: osx
       env: PYTHON_VERSION=3.7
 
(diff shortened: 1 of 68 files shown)


QuLogic:
You should add minimum Python to `setup.py` too, [`python_requires`](https://github.com/matplotlib/matplotlib/blob/90fba030f534b3d7068e536fcfc4a2dc8d459eeb/setup.py#L257).

MarkWieczorek:
Does this look ok to you? I set the minimum python version to 3.4 (which I know works with the travis tests). If there is a reason to use a lower version, I can change this

MarkWieczorek:
I changed the minimum version of python to 3.5 because (1) numpy claims to require >=3.5, and (2) 3.4 has already hit end-of-life.

QuLogic:
You may want to move that initial check to above all other imports as well (except `sys`), so that it's checked as early as possible and you are free to do whatever you like after.

PR #198, opened 2019-10-07
Clarify docs in spharm_functions.py

jammm:
I think this might make it a bit clearer to understand. If this is not acceptable, feel free to close this PR.

Thanks!

### DIFF
--- a/pyshtools/expand/spharm_functions.py
+++ b/pyshtools/expand/spharm_functions.py
@@ -37,9 +37,9 @@ def spharm(lmax, theta, phi, normalization='4pi', kind='real', csphase=1,
     lmax : integer
         The maximum degree of the spherical harmonic functions to be computed.
     theta : float
-        The colatitude in degrees.
+        The colatitude in degrees. Use radians if 'degrees' is set to False.
     phi : float
-        The longitude in degrees.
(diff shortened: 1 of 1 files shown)


pyshtools/expand/spharm_functions.py
```diff
@@ -218,7 +218,7 @@ def spharm_lm(l, m, theta, phi, normalization='4pi', kind='real', csphase=1,
         Condon-Shortley phase of (-1)^m will be appended to the spherical
         harmonic functions.
     degrees : optional, bool, default = True
-        If True, colat and phi are expressed in degrees.
+        If True, 'theta' and 'phi' are expressed in degrees.
```
MarkWieczorek: can you use back ticks for consistence (and also for use with the web documentation)?

MarkWieczorek:
Hi. This looks good to me, but could you base this PR on the develop branch? We only use master for releases.

Also, could you use backticks for all variables that are used by the function? This is not only for consistency, but it makes a difference in how this is rendered for the html documentation. Thanks!

pyshtools/expand/spharm_functions.py
jammm: Ah, I missed that. Thanks!

jammm:
Closing PR and making a new one for `develop` branch.

PR #213, opened 2020-01-24
add netcdf io for SHCoeffs class

Nasuyo:
Hi,

I felt the lack of a fast way of saving the coefficients. For the SHGrid, "to_netcdf()" exists already. For the coefficients, netcdf is of course not the perfect format, but I believe it is still a very good addition to the existing formats due to its saving time and (compared to the textfiles) low disk space (~1/3).

This pull request is not final, I just wanted to know if you are generally interested in the 2 functions I included.

- [x]  Save coefficients
- [x]  Load coefficients
- [x]  Add functions to class docstring
- [x]  Add max degree parameter for loading
- [x]  Add csphase parameter for loading
- [x]  Add normalization parameter for loading

Best Regards,
Stefan


MarkWieczorek:
Hi. I don't have a problem adding this if you think it is useful. In terms of saving space

1. You could save the coefficients as a numpy array (but you would lose some meta data).
2. We could add an option to save and read the coefficients as a binary file, instead of text as it is now.

In terms of netcdf, I agree with what you note above.

1. We should save/read the `csphase` and `normalization` constants.
2. We should add an `lmax` option that allows to read only a part of the file.

Otherwise, no problems!

MarkWieczorek:
Could you base this PR on the develop branch? We only publish to master for releases...

MarkWieczorek:
This PR is modifying the several unintended files (such as the makefile and setup.py). I think that this is because you are working on the master branch on your local repo, but trying to merge this into develop here. It would be easiest for me if you could base your changes off of develop so that we can easily merge into develop.

I am aiming for a new release on about March 13. It would be great to include this!

Nasuyo:
I reopened this PR in #217

PR #215, opened 2020-02-05
Center of mass and inertia tensor calculation

ioshchepkov:
This PR adds the following methods  to `SHGravCoeffs` class:

- `center_of_mass` for the center of mass calculation;
- `inertia_tensor()` for the moment of inertia tensor.

Since the latter method requires dynamical flattening constant H it is also added for the Earth from the IERS Conventions 2010.

The `read_icgem_gfc()` function is extended with the option `encoding`. It turned out that some models in ICGEM are not in UTF-8.


MarkWieczorek:
While we are doing this, maybe we should add a `center_of_figure()` method to the `SHCoeffs`.

ioshchepkov:
@MarkWieczorek I'm not very familiar with that, do you have any references for COF calculation? I mean, is it used anywhere other than topography?

pyshtools/shclasses/shgravcoeffs.py
```diff
@@ -806,6 +810,89 @@ def mass(self):
...
+
+        Returns
+        -------
+        x, y, z : float
+            Cartesian coordinates of the center of mass, in metres.
+        """
+        coeffs = self.convert(normalization='unnorm',
+                csphase=1, lmax=1).coeffs
+
+        x_cm = coeffs[1, 1, 1] * self.r0
```
MarkWieczorek: I think that the formulas for x_cm and y_cm are switched. coeffs[0,1,1] is C11 and coeffs[1,1,1] is S11

ioshchepkov: Oops, I agree!

MarkWieczorek:
I'll take care of the center-of-figure calculation when this PR is merged. This will be `SHCoeffs.centroid()` and is simply the center-of-mass coordinates for a uniform density body.

MarkWieczorek:
I am aiming to put out a new release on about March 13, and would like to include this PR. There are only a couple minor documentation issues that need to be resolved:

First (required), we should define in the docstring what the dynamical-flattening/precession-constant is (I always forget). I believe that this is just [C-0.5(A+B)]/C.
Second (optional), it could be useful to note in the docstring that this routine is based on the assumption that Ixx, Iyy, and Izz are closely aligned with A, B and C. I believe that this works because Ixx, Iyy and Izz are much larger than the off diagonal terms. I suspect that there is an exact complicated analytical solution to this problem, but I have never had the need for that level of accuracy.
Third (optional, very minor): For consistence, I would use "meters" instead of "metres".

PR #217, opened 2020-02-25
Add netcdf io to shcoeffsgrid.py

Nasuyo:
For simplicity I just reopened PR #213 .

Only 1 file changed, there should be no problems with merging anymore.

Best,
Stefan

### DIFF
--- a/pyshtools/shclasses/shcoeffsgrid.py
+++ b/pyshtools/shclasses/shcoeffsgrid.py
@@ -642,7 +642,7 @@ class SHCoeffs(object):
 
         Usage
         -----
-        x = SHCoeffs.from_netcdf(filename)
+        x = SHCoeffs.from_netcdf(filename, [lmax, normalization, csphase])
 
         Returns
         -------
@@ -684,9 +684,6 @@ class SHCoeffs(object):
                 "csphase must be 1 or -1. Input value was {:s}"
                 .format(repr(csphase))
                 )
-    
-        if not filename[-3:] == '.nc':
-            filename += '.nc'
         
         ds = _xr.open_dataset(filename)
         lmaxout = ds.dims['degree'] - 1
@@ -811,7 +808,7 @@ class SHCoeffs(object):
         
         Usage
         -----
-        x.to_netcdf([filename, title, description])
+        x.to_netcdf(filename, [title, description])
         
         Parameters
         ----------
@@ -822,9 +819,6 @@ class SHCoeffs(object):
         description : str, optional, default = ''
             Description of the data.
         """
-        
-        if not filename[-3:] == '.nc':
-            filename += '.nc'
         
         ds = _xr.Dataset()
         ds.coords['degree'] = ('degree', _np.arange(self.lmax+1))
--- a/pyshtools/shclasses/shcoeffsgrid.py
+++ b/pyshtools/shclasses/shcoeffsgrid.py
@@ -44,6 +44,7 @@ class SHCoeffs(object):
         x = SHCoeffs.from_random(powerspectrum)
         x = SHCoeffs.from_zeros(lmax)
         x = SHCoeffs.from_file('fname.dat')
+        x = SHCoeffs.from_netcdf('ncname.nc')
         x = SHCoeffs.from_cap(theta, lmax)
 
     The normalization convention of the input coefficents is specified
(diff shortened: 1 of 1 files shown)


pyshtools/shclasses/shcoeffsgrid.py
```diff
@@ -629,6 +633,91 @@ def from_cap(self, theta, lmax, clat=None, clon=None, normalization='4pi',
...
+    
+    @classmethod
+    def from_netcdf(self, filename, lmax=None, normalization='4pi', csphase=1):
+        """
+        Initialize the class with spherical harmonic coefficients from a 
+        netcdf file.
+
+        Usage
+        -----
+        x = SHCoeffs.from_netcdf(filename)
```
MarkWieczorek: add 

, [lmax, normalization, csphase])

to usage string.

pyshtools/shclasses/shcoeffsgrid.py
```diff
@@ -629,6 +633,91 @@ def from_cap(self, theta, lmax, clat=None, clon=None, normalization='4pi',
...
+                .format(repr(normalization))
+                )
+
+        if csphase != 1 and csphase != -1:
+            raise ValueError(
+                "csphase must be 1 or -1. Input value was {:s}"
+                .format(repr(csphase))
+                )
+    
+        if not filename[-3:] == '.nc':
```
MarkWieczorek: I think that this is a good idea if `filename` does not have an extension. However, if you provide an extension like `filename.netcdf` then I don't think we need to append another extension `.nc`.

pyshtools/shclasses/shcoeffsgrid.py
```diff
@@ -629,6 +633,91 @@ def from_cap(self, theta, lmax, clat=None, clon=None, normalization='4pi',
...
+        ds = _xr.open_dataset(filename)
+        lmaxout = ds.dims['degree'] - 1
+            
+        c = _np.tril(ds.coeffs.data)
+        s = _np.triu(ds.coeffs.data, k=1)
+        s = _np.vstack([s[-1], s[:-1]])
+        s = _np.transpose(s)
+        if isinstance(lmax, int):
+            c, s = c[:lmax+1, :lmax+1], s[:lmax+1, :lmax+1]
+            lmaxout = lmax
```
MarkWieczorek: lmaxout in this line is never used later.

pyshtools/shclasses/shcoeffsgrid.py
```diff
@@ -717,6 +806,40 @@ def to_file(self, filename, format='shtools', header=None, **kwargs):
...
+        ----------
+        filename : str
+            Name of the output file.
+        title : str, optional, default = ''
+            Title of the dataset
+        description : str, optional, default = ''
+            Description of the data.
+        """
+        
+        if not filename[-3:] == '.nc':
```
MarkWieczorek: Same comment as above: I think that if the user provides a filename with an extension that is not `.nc` that we shouldn't append the extension `.nc`.

pyshtools/shclasses/shcoeffsgrid.py
```diff
@@ -717,6 +806,40 @@ def to_file(self, filename, format='shtools', header=None, **kwargs):
             raise NotImplementedError(
                 'format={:s} not implemented.'.format(repr(format)))
 
+    def to_netcdf(self, filename, title='', description=''):
```
MarkWieczorek: I think that it would be a good idea to be able to specify an optional `lmax` to read, like the `from_netcdf()` method above.

pyshtools/shclasses/shcoeffsgrid.py
Nasuyo: Done.

pyshtools/shclasses/shcoeffsgrid.py
Nasuyo: Alright, I deleted the lines. Generally I also think it is a good idea, but it would be inconsistent with the other functions anyways.

pyshtools/shclasses/shcoeffsgrid.py
Nasuyo: It is used in line 702.

pyshtools/shclasses/shcoeffsgrid.py
Nasuyo: Deleted as well.

pyshtools/shclasses/shcoeffsgrid.py
Nasuyo: Generally I agree, but I think this should be done consistently for all to_xxx() functions then. I'll leave it like this if that's fine.

PR #252, opened 2020-09-14
parallel make

### LINKED ISSUE #196: `make -j fortran` (parallelized) is not stable
When building the fortran components of SHTOOLS from the makefile, compilation will sometimes give an error when using `make -j fortran` (the `-j` option compiles the files in parallel). 

The problem is that some of the `.mod` files are required before compiling a given piece of code. In the makefile (`src/Makefile`), I have explicitly tried to list all the dependencies explicitly using 

```
ftypes.mod : ftypes.o
SHTOOLS.mod : SHTOOLS.o ftypes.mod
FFTW3.mod : FFTW3.o ftypes.mod
PlanetsConstants.mod : PlanetsConstants.o ftypes.mod
...
```
However, sometimes (not always) when compiling a file like `PlanetsConstants.f95`, the compiler will complain that `ftypes.mod` does not exist.

It is possible that this is a problem with `make` (I am using the system `make` on macOS).

As a temporary solution with the macOS brew package manager (which defaults to parallel compilation), I have explicitly required that brew compile these files sequentially (see https://github.com/SHTOOLS/homebrew-shtools/issues/4).

I note that this problem occurs about 1/6th of the time (or less) when compiling. If this error occurs, it is usually only necessary to try again.

rainbowsend:
closes  #196

errors like this
```
PlanetsConstants.f95:11:9:

   11 |     use ftypes
      |         1
Fatal Error: Cannot open module file ‘ftypes.mod’ for reading at (1): No such file or directory
compilation terminated.
```
do not longer occur when using parallel make. I have called `make clean; make -j 16` 25 times without  such an error message


MarkWieczorek:
This is great! The makefile is much cleaner too. I'm just going to merge this. I am probably going to put out a 4.7.1 release "soonish" because I need to make a couple of other changes to the makefiles to get shtools into the homebrew-core and macports package managers.

PR #256, opened 2020-09-23
Minor changes of cpp makefiles

MarkWieczorek:
This PR makes a few very minor changes to the Makefiles:

* All compiler options for the cpp example are passed to the cppmakefile
* The cpp make file now has a run target.


MarkWieczorek:
@rainbowsend Are you ok with these small changes to the makefiles ?

Note, this currently does not compile on macOS with homebrew gcc-10. In addition to using the following make command:
```
make cpp-tests-no-timing CXX='g++-10' LAPACK='-framework Accelerate' BLAS=''
```
it is necessary to edit the `examples/cpp/makefile` and modify LIB accordingly
```
LIB = -L $(LIBPATH) -l$(LIBNAME) -L /usr/local/lib/gcc/10/ -lgfortran -ldl $(FFTW) $(LAPACK) $(BLAS)
```

I need to do some digging to figure out what homebrew does with gcc.

rainbowsend:
Yeah this is running on my machine as expected. However, there are two makefiles in /examples/cpp/. One with capital M and one with small m. I forgot to include the deletion of `makefile` the in last commit. Since, you have edited this one, we could  rename `makefile` to `Makefile` (deleting the original `Makefile`)

rainbowsend:
is adding `-stdlib=libc++` to the CXXFLAGS solving the problem on macOS?

MarkWieczorek:
No, that doesn't help (or even work....). I think that this is a homebrew problem and I'm discussing this with someone over there.

MarkWieczorek:
https://github.com/Homebrew/discussions/discussions/11

Issue #300, opened 2021-05-29
Inaccuracy in conversion from complex grid to SHCoeff


mreineck:
The following snippet generates SHCoeffs, converts them to a grid and back again, and compared the result to the input. The resulting number should be very small (around 1e-15), but in some cases like the presented one it is much larger.

```
import numpy as np
import pyshtools as pysh

lmax=15
csphase=1
grd="DH"
normalization="4pi"
extend=True
degrees = np.arange(2*lmax+1, dtype=float)
degrees[0] = np.inf
power = degrees**(-2)

clm = pysh.SHCoeffs.from_random(power, seed=12345, kind='complex')
clm = clm.convert(normalization=normalization,
                  csphase=csphase,
                  lmax=lmax)

grid = clm.expand(grid=grd,extend=extend)
cilm = grid.expand(normalization=normalization,
                  csphase=csphase)

print(np.max(np.abs(cilm.to_array()-clm.to_array())))
```

This inaccuracy only happens when
- the grid is "DH" or "DH2"
- lmax is odd
- the transforms are complex-valued

As far as I can tell, there is some problem in the `grid.expand` call when computing the coefficient for (lmax, lmax), but I didn't analyze this any further.

MarkWieczorek:
This is fixed in this PR. I will merge this into develop within a week after I do a bit more testing.

https://github.com/SHTOOLS/SHTOOLS/pull/303

PR #303, opened 2021-06-06
Fix 2 bugs with complex spherical harmonic transforms

MarkWieczorek:
This PR fixes 2 bugs related to complex spherical harmonic transforms.

The first regards the Fortran routine `SHExpandDHC` and hence affects both shtools and pyshtools. For complex grids, the **last** coefficient `coeffs[1, lmax, lmax]` was in error only when `lmax` was odd. The origin of this is a missing (-1)^m factor that should have been applied to a term.

The second regards the `SHGrid.expand()` routine for complex data with grid type `DH2`. The origin of this error is that the parameter `sampling=2` was not passed to the Fortran routine. The spherical harmonic coefficients for this case were all wrong, and this should have been obvious by simple inspection.

Fixes: https://github.com/SHTOOLS/SHTOOLS/issues/300

### DIFF
--- a/pyshtools/shclasses/shgrid.py
+++ b/pyshtools/shclasses/shgrid.py
@@ -2189,7 +2189,8 @@ class DHComplexGrid(SHGrid):
 
         cilm = _shtools.SHExpandDHC(self.data[:self.nlat-self.extend,
                                               :self.nlon-self.extend],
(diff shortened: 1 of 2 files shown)


mreineck:
Great, thanks a lot!

PR #306, opened 2021-07-16
patched a floating point error-caused bug (arccos of value >= 1)

nephanth:
Hi !
I just encountered a small bug in the python code, so I wrote a simple fix, if you want to merge it

Basically on rare cases, in `shgrid.from_cap`, when calculating `dist`, floating-point calculation errors could give a result slightly superior to 1 (I had `1.0000000000000002`). In that case, the `arccos` at the following line will output `NaN`, and may raise an exception (if `numpy` is configured to do so, see [numpy.seterr](https://numpy.org/doc/stable/reference/generated/numpy.seterr.html) ).

So what I did was cap that value to `1.` (see the changelog)

### DIFF
--- a/pyshtools/shclasses/shgrid.py
+++ b/pyshtools/shclasses/shgrid.py
@@ -362,8 +362,9 @@ class SHGrid(object):
             sinlat = _np.sin(lats[i])
             for j in range(0, temp.nlon):
                 dist = coslat * (x * coslon[j] + y * sinlon[j]) + z * sinlat
-                if dist <= costheta: # ie. _np.arccos(dist) <= theta 
+                if dist >= costheta: # ie. _np.arccos(dist) <= theta 
                     #since 0 <= theta <= pi/2 and 0 <= dist <= 1
+                    #so cos is decreasing
                     temp.data[i, j] = 1.
 
         return temp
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Thanks. I will look into this the first week of august.

mreineck:
Personally, I'm always using this formula to determine the angle between two vectors:

`alpha = atan2 (crossprod(v1,v2).Length(), dotprod(v1,v2))`

Advantages:
- not really slower than other approaches
- extremely accurate even for angles near 0 and pi (which is not true for anything involving `acos`)
- works for non-normalized vectors
- does not risk calling `acos(1.+epsilon)` and similar

For me this has proven so convenient that I even use it if I have (theta, phi) tuples as input - I just convert them to 3-vectors first.

MarkWieczorek:
If you want to change the routine to use this approach, that would be ok with me. I don't really have a strong opinion one way or the other.

mreineck:
Just had another look ...
The pragmatic solution might be to just change
` if _np.arccos(dist) <= theta:`
into
` if  dist >= np.cos(theta):`
What do you think?
(For extra points, `np.cos(theta)` could be precomputed.)

MarkWieczorek:
Maybe this is the best (and fastest) solution! And yes, I would just define something like `costheta = np.cos(theta)` before loop so that this doesn't get computed everytime.

MarkWieczorek:
@nephanth Did you want to make the change using `if dist >= np.cos(theta):` since you found this problem? If its not done on monday, I'll probably just do this myself.

nephanth:
done!

PR #330, opened 2022-01-26
Fghorow patch 1

fghorow:
As requested, also changed URL for sha.grgm1200b_rm1_1e0_sigma Lunar gravity potential file from NASA Goddard. The SHA256 checksums match between the old and new URL's files.

**Reminders**


MarkWieczorek:
I noticed that the you changed the `tag_prefix` in the file `setup.cfg`. Is there a reason for this? I believe that this was necessary with previous releases, but maybe things have changed with a new version of versioneer.

MarkWieczorek:
The updated urls have been merged into develop in my last PR. We will be putting out a new release in 1 or 2 weeks.

PR #337, opened 2022-03-25
convert input arrays if necessary

mreineck:
**Reminders**


mreineck:
I don't think the test failures are related to my changes; no idea what's going on.

PR #341, opened 2022-03-31
Make ducc the default backend

MarkWieczorek:
ducc was set as the default backend in a previous PR, but this PR improves the handling of the case where ducc0 is not installed.

### DIFF
--- a/docs/pages/mydoc/backends.md
+++ b/docs/pages/mydoc/backends.md
@@ -24,8 +24,8 @@ The following software packages are supported as numerical backends to *pysthool
 
 | Class name | Description |
 | ---------- | ----------- |
-| [shtools](index-fortran.html) (default) | Spherical Harmonic Tools (Fortran 95, installed as part of pyshtools) |
-| [ducc](https://gitlab.mpcdf.mpg.de/mtr/ducc) | Distinctly Useful Code Collection (C++17) |
+| [shtools](index-fortran.html) (fallback) | Spherical Harmonic Tools (Fortran 95, installed as part of pyshtools) |
+| [ducc](https://gitlab.mpcdf.mpg.de/mtr/ducc) (default) | Distinctly Useful Code Collection (C++17) |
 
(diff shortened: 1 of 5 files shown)


mreineck:
Looks good, only one thing: should we check for ducc version >= 0.15 in `backend_module` as well? It's posible that I forgot to do that ...

mreineck:
Actually we test everything properly when importing `ducc0_wrapper` ... so we do a bit too much work in `select_preferred_backend`, I think. I'll prepare a patch.

mreineck:
Hmm, preparing a patch against `MarkWieczorek:develop` seems tricky, so I'd rather wait until this is merged...

mreineck (approved):

MarkWieczorek:
ok. done!

mreineck:
Ah, sorry! That makes sense because this code calls the "low-level" expand routine.

MarkWieczorek:
Actually, this is a problem for all of the timing scripts. The problem is the following:

1. Everything works when using the `SHGrid `and and `SHCoeffs` class.
2. However, the original low-level SHTOOLS routines can be found in two places, in the `pysh.shtools.MakeGridDH()` submodule, or categorized in submodules like `pysh.expand.MakeGridDH()`
3. The timing scripts call `pysh.expand.MakeGridDH()`

Problem 1: in the top level namespace, there is a module with all the "shtools" routines, but there isn't a similar module for the "ducc" routines. The ducc routines are located here: `pysh.backends.ducc0_wrapper`. Should we create a convenience link in the top level namespace to `pysh.backends.ducc0_wrapper`?

Problem 2: If you select the ducc backend, the routines in submodules like `expand` will still point to the old shtools routines. This could be fixed by having the `select_preferred_backend()` routine create links to the correct routines in each of the submodules. I don't know if there is an elegant way to do this.

Problem 3: when pyshtools was first created, it didn't have any classes, and everything was done using the low-level functions. Now, people probably don't use the low level functions. Should we rethink how the entire namespace is orgnaized in order to de-emphasize the low-level routines?

Any ideas? Not really sure how best to proceed...

PR #342, opened 2022-03-31
Simplify backend management

mreineck:
This keeps track of the available backends in a dictionary, which allows removing a lot of redundant code.

**Reminders**


MarkWieczorek:
much better! Also, did you see my comment about conda and osx-arm64 packages in the chat room? ducc0 is the only dependency here that doesn't have an osx-arm64 conda package. I'm not sure how to make one, but you might just need to re-render at the conda-forge feedstock.

PR #344, opened 2022-04-01
Enable backend support in a few global functions

mreineck:


mreineck:
@MarkWieczorek I think this covers all functions that are implemented in `ducc`. What do you think about this appoach?

mreineck:
Should I do something analogous in `shtools/__init__.py`?

mreineck:
`rotate/__init__,py` is a tricky candidate, since the precomputation is no longer necessary. I don't think we can do much there.

MarkWieczorek:
I tested one of these, and it seems to work fine. I don't have a problem with the approach.

For your question about the `shtools` module: I would just leave this alone as it is for the fortran routines. I would eventually like to remove this module all together as it is the same thing as `backends.shtools`. The problem is that the `shtools` module is hardcoded in a lot of places, and this will all need to be cleaned up.

Nevertheless, I realized three (mostly cosmetic) things when looking at your changes.

First, there is a directory `pyshtools/wrappers` with an empty init file in it. As far as I can tell we don't use this anymore, and it is probably just an orphan from when were were making the backend module. If you agree, could you delete this?

Secondly, I realized that the ducc routines in `pyshtools/backends/ducc0_wrappers` don't have doc strings.

Lastly, your changes above clobber the pre-existing docstrings.

Adding docstrings to the shtools functions was done by reading the documentation from a text file and then manually setting  `__doc__`. This is done in the init file for the `shtools` module. I'll need to look at this later to figure out how this was done in practice, but we should probably use the same approach to set the docstrings in `pyshtools/backends/ducc0_wrappers`, and then make sure that we copy them to the new functions you define.

MarkWieczorek:
I tested one of these, and it seems to work fine. I don't have a problem with the approach.

For your question about the `shtools` module: I would just leave this alone as it is for the fortran routines. I would eventually like to remove this module all together as it is the same thing as `backends.shtools`. The problem is that the `shtools` module is hardcoded in a lot of places, and this will all need to be cleaned up.

Nevertheless, I realized three (mostly cosmetic) things when looking at your changes.

First, there is a directory `pyshtools/wrappers` with an empty init file in it. As far as I can tell we don't use this anymore, and it is probably just an orphan from when were were making the backend module. If you agree, could you delete this?

Secondly, I realized that the ducc routines in `pyshtools/backends/ducc0_wrappers` don't have doc strings.

Lastly, your changes above clobber the pre-existing docstrings.

Adding docstrings to the shtools functions was done by reading the documentation from a text file and then manually setting  `__doc__`. This is done in the init file for the `shtools` module. I'll need to look at this later to figure out how this was done in practice, but we should probably use the same approach to set the docstrings in `pyshtools/backends/ducc0_wrappers`, and then make sure that we copy them to the new functions you define.

mreineck:
> First, there is a directory pyshtools/wrappers with an empty init file in it. As far as I can tell we don't use this anymore, and it is probably just an orphan from when were were making the backend module. If you agree, could you delete this?

Sure! I also think that this is no longer needed.

> Secondly, I realized that the ducc routines in pyshtools/backends/ducc0_wrappers don't have doc strings.

I can add some, but they will basically be verbatim copies of the `shtools` ones (since the functions are supposed to do exactly the same thing after all :-) But I can document that some parameters (like Gauss-Legendre zeros and weights) will be ignored.

Concerning the clobbered docstrings: I'm not sure if I have enough Python knowledge to fix this, but I'll try to help if necessary.

mreineck:
OK, there will actually be some differences in the docstrings, since they contain quite a few technical details... this might take a few days.

If that turns out to take too long, my suggestion is not to hold back the release just for this...

mreineck:
It seems that the parameter names in my functions don't match those in the shtools docstrings (e.g. "w" vs "weights", "zero" vs. "zeros"). Where can I find the canonical names to use?

mreineck:
I just noticed that the patch in the current form breaks backwards compatibility because of this name mismatch. I'll try to fix this.

MarkWieczorek:
If you have any questions about the variable names used by python, I would take a look at `src/PythonWrapper.f95`. This is the wrapper for the f95 files. As long as things work, I'm not too concerned about this....

mreineck:
Thanks, I think I found everything I need now!

> As long as things work, I'm not too concerned about this....

There are corner cases where the actual parameter names are significant. Imagine some use code calling
`SHExpandGLQ(..., w=<something>,...)`; if this call is rerouted to the `ducc` backend, it would have failed, because the corresponding function there did not have a parameter called `w`, only `weights`.

Anyway I think this is now ready to merge. `make check` complains about long lines in the new docstrings, but I'm not sure it's worth reformatting hundreds of lines; `black` seems to be happy.

MarkWieczorek:
I think I have a solution for changing the docstrings. In `backends.py`, I created a function called `write_backend_docstrings` and then call this in the `select_preferred_backend` routine.

```
def select_preferred_backend(backend="ducc", nthreads=None):
    """
    Select the preferred backend module used for the spherical harmonic
    transforms in pyshtools.

    Usage
    -----
    select_preferred_backend([backend, nthreads])

    Parameters
    ----------
    backend : str, optional, default = 'ducc'
        Name of the preferred backend, either 'shtools' or 'ducc'.
    nthreads : int, optional, default = 1
        Number of threads to use for the 'ducc' backend. Setting this parameter
        to 0 will use as many threads as there are hardware threads on the
        system.
    """
    global _preferred_backend
    backend = backend.lower()
    if backend in _available_backends:
        _preferred_backend = backend
        if backend == "ducc" and nthreads is not None:
            ducc0_wrapper.set_nthreads(nthreads)
        write_backend_docstrings(backend=backend)
    else:
        print("Requested backend '{}' not available.".format(backend))
        raise RuntimeError


def write_backend_docstrings(backend=None):
        from .. import expand
        expand.MakeGridGLQ.__doc__ = backend_module(backend=backend).MakeGridGLQ.__doc__
        expand.MakeGridDH.__doc__ = backend_module(backend=backend).MakeGridDH.__doc__
```
Of course, we would have to add all the modules that get changed in `write_backend_docstrings`. We also would need to call this function once when initializing pyshtools in order to set the docstrings in the beginning (I'm less sure where to put this call).

What do you think of this approach? If you want to take care of this, great. Otherwise, just let me know and I'll work on it.

mreineck:
> And I also found a solution to the line length warnings.

Ah, great, that is extremely useful!

Concerning the manipulation of docstrings, I'll need a bit of time to understand how all of this works :-) Will get back to you.

mreineck:
There is also a completely different way to switch the functions depending on the backend: they could just be injected into the `expand` module by adding something like this in `expand/__init__.py`

```
def inject_backend_specific_functions():
    mod = backend_module()
    global SHExpandGLQ
    SHExpandGLQ = mod.SHExpandGLQ
    [...]
```

Whenever the preferred backend changes, this function is called and makes the necessary replacements. This would also take care of the docstrings, I think.

I don't know whether this is a nice solution; I'm not enough of a Python expert to judge that.

mreineck:
Sorry, I don't think I will have time to address the docstring issue properly in the short term. If you have a solution you like, please go ahead!

The remaining overlong lines outside of the docstrings should be fixed now.

MarkWieczorek:
I'm thinking that your new approach might be a little cleaner. All we need to do is call `inject_backend_specific_functions()` from within `select_preferred_backend()`, and then in `backends/__init__.py` just call `select_preferred_backend()` at the end of the file in order to initialize everything at startup. It would probably be easier for your to do this, but if you wanted, I could try making a PR on your private repo. I am in no immediate hurry, as I am trying to get osx-arm64 conda packages for `ducc0` and `pypandoc` before making a release. If there is no progress on those by April 15, I'll probably just go ahead with making a release anyway.

As for the rotate functions, I'm not sure if I see what the problem is. For the shtools routines, we don't need to precompute anything: if you don't specify the dj rotation maxtrix, this will get computed internally. I don't think that ducc0 has an issue with precomputation, but maybe I didn't look close enough. In any case, if we add this, we will need to make a separate `inject_backend_specific_functions()` for the `rotate` module.

mreineck:
I'll see what I can do over the weekend!

> As for the rotate functions, I'm not sure if I see what the problem is. For the shtools routines, we don't need to precompute anything: if you don't specify the dj rotation maxtrix, this will get computed internally.

You are right; I had forgotten that point.

mreineck:
> As for the rotate functions, I'm not sure if I see what the problem is. For the shtools routines, we don't need to precompute anything: if you don't specify the dj rotation maxtrix, this will get computed internally.

After double-checking, I think this is only true for the rotation functions inside the `SHCoeffs` class,but not in the standalone function. At least `pyshtools.pyf` does not declare `dj`  optional for `SHRotateRealCoef`.

One more thing: I'm trying to figure out the role of `SHRotateCoef` ... it seems like a variant of `SHRotateRealCoef` that takes the coefficients in a slightly different format. Is this still in active use?

MarkWieczorek:
I think that you are right about the dj matrix with `rotate.SHRotateRealCoef`. However, with the way we are re-linking the functions to the backend, I don't think that this matters too much. If someone provided the dj matrix to ducc0, it would just get ignored. The only thing that is worrying to me is that dj is mandatory for shtools, but optional and ultimately ignored by ducc, and this might need to be explained better in the docs. For example, in the shtools docstring we could just say (Note that even though this variable is required when using the shtools backend, it is not required with other backends.)

As for `SHRotateCoeff`, I need to emphasize that this is the only code that I didn't write myself. I have always wanted to re-write this, but never got around to it. It's my recollection that this routine makes use of a non-default phase convention and normalization, and that it uses complex coefficients to rotate a strictly real function. I think that the coefficients are also packed in a more efficient manner that doesn't have zeros for half the matrix. The `SHRotateRealCoeff` is simply just a fortran wrapper that makes it easier to use.

mreineck:
OK, thanks! `SHRotateRealCoef` is now also injected.

Concerning storage of harmonic coefficients, my own experience is that the most important aspect is to have `l` as the fastest varying index, since in practice all algorithms that require high performance will iterate over `l` in their innermost loops.
`ducc` also stores coefficients for real-valued maps as complex numbers. This is not a perfect solution, but compact,real-valued storage schemes come with their own set of drawbacks ...
If you consider updating the storage scheme at some point in the future, I'd be happy to discuss this further!

MarkWieczorek:
The failing checks is because of a problem with a new version of setuptools (62.1.0). Everything works fine using 62.0.0 or lower. I am just going to merge this now. If setuptools doesn't fix this before our release (I am aiming for april 18), then I will pin the version of setuptools to avoid this problem.

https://github.com/pypa/setuptools/issues/3263

PR #360, opened 2022-09-06
Fix _expand_coord in SHGravRealCoeffs and SHMagRealCoeffs

kohlerjl:
SHGravRealCoeffs.expand and SHMagRealCoeffs.expand are producing incorrect results when called with a list of latitude and longitude points, because the same radius is being used to calculate all the points. This pull request fixes the radius calculation to use the latitude of the current point in the iteration only


MarkWieczorek:
Thanks for fixing this!

PR #375, opened 2023-03-07
fix import of pkg_resources

k-dominik:
While trying to [fix the conda recipe for arm64](https://github.com/conda-forge/pyshtools-feedstock/pull/37), I ran into an issue for all platforms with updated setuptools versions.

`setup.py` was previously relying on `pkg_resources` being imported from `setuptools`. That broke in
https://github.com/pypa/setuptools/commit/6050634818943befefe3a85a12503b6d8a1e8106

**Reminders**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ try:
     import pypandoc
     rst = pypandoc.convert_file('README.md', 'rst')
     long_description = rst.split('\n', 5)[5]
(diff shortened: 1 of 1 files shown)


MarkWieczorek:
Thanks for the fix! I'm going to spend the next 2 weeks going though the backlog of issues and will get to your PR for the ARM conda issue soon...

PR #378, opened 2023-03-20
Move man pages to section 3

opoplawski:
Library calls really belong in section 3.


MarkWieczorek:
Thanks for the PR. You are right that section 1 probably wasn't the best choice. This build is failing because of this pkg_resources issue : https://github.com/SHTOOLS/SHTOOLS/pull/375 . I'm just going to accept this anyways as it looks pretty straightforward.

PR #379, opened 2023-04-05
Allow passing test input data from command line

QuLogic:
When building with Meson, the build will _always_ be in an out-of-source build directory, so tests cannot rely on relative paths like these do to find their data.

This PR reads the data path from the command line, which allows testing from any build directory, no matter where it is, as the path to the source directory can be passed.

I also had to modify the an additional input file to some of the Fortran tests, in order to supply test options directly, instead of as data on standard input.

**Reminders**

### DIFF
--- a/examples/cpp/test.cpp
+++ b/examples/cpp/test.cpp
@@ -26,7 +26,7 @@ int
 main(int argc, char** argv)
 {
 
-  std::string infile = "../ExampleDataFiles/MarsTopo719.shape";
+  std::string infile;
 
   int lmax = 15;
   int cilm_dim = lmax + 1;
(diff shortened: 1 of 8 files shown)


MarkWieczorek:
I really appreciate the work you are doing on this, and apologize for the backlog of PRs and issues. I realize that this is starting to become urgent, and my plan it to devote significant time to this starting April 17. Things have just been very busy for me lately...

MarkWieczorek:
Is this PR incorporated into this larger Meson PR? https://github.com/SHTOOLS/SHTOOLS/pull/380

If so, can we delete this? If not, could we rebase it on the most recent commit to see if the CI checks pass?

QuLogic:
It is part of #380, but I went and rebased it anyway. I think it's easier to review both of them separately.

MarkWieczorek:
Looks good !

PR #380, opened 2023-04-05
Add a Meson build system

QuLogic:
You can either build the library for installation or for a Python package, but I've mostly tuned it for the Python package. 

This is based on #369 and #379.

**Reminders**


pyshtools/meson.build
```diff
@@ -0,0 +1,51 @@
+python_sources = [
+  '__init__.py',
+  '_version.py',
+]
+
+pymod = import('python')
+python = pymod.find_installation(modules: ['numpy'])
```
eli-schwartz, line 7: If you bumped the minimum required meson version up to 0.64.0 then you could pass `pure: false` as a kwarg to find_installation and not have to repeat it everywhere else (also not have to worry if you forget it).

QuLogic: The minimum was pretty old; I can probably bump that up to 1.0.0 at this point.

QuLogic:
So apparently, I must have tried to set this up outside a conda environment, as there were a few tweaks to search paths that needed fixing here. But now it builds on CI. I did try out Windows builds as well earlier, but those seem to need some finessing to get conda-forge compilers correct, so I'd rather not do that here.

Now that I can build and install this, I see there are (at least) two outstanding issues:
- [ ] For versioning, I simply made the versioneer `_version.py` 'static'; this means it can synthesize a version out of git, but isn't really correct with regard to tags. I'll probably move that to some Meson substitution if that works.
- [ ] I didn't notice the special docstring builder earlier, but now see all its errors when importing, so I'll have to wire that in to the Meson build.

pyshtools/meson.build
```diff
@@ -0,0 +1,51 @@
+python_sources = [
+  '__init__.py',
+  '_version.py',
+]
+
+pymod = import('python')
+python = pymod.find_installation(pure: false, modules: ['numpy'])
```
rgommers, line 7: The `modules` argument is actually a little dangerous, for at least two reasons:
- the check is done by importing the module, so it doesn't work if there's no importable Python package or if the import name doesn't match the package name
- it may interfere with cross-compiling

I'll follow up on that upstream, but I think it's best left out here.

pyshtools/meson.build
```diff
@@ -0,0 +1,51 @@
...
+
+# NumPy include directory
+incdir_numpy = run_command(python,
+  [
+    '-c',
+    'import os; os.chdir(".."); import numpy; print(numpy.get_include())'
```
rgommers, line 13: This has issues with in-tree virtualenvs, which is an annoying problem. You may want to take over https://github.com/scipy/scipy/blob/1a3021a08e5898b52d7b98f7b1b21ae4a6306484/scipy/meson.build#L56-L82 here.

rgommers (commented):
This looks quite good, just two suggestions for robustness.

MarkWieczorek:
I am going to try to put out a new minor release v4.10.2 in the next few days, and then try to switch to meson for v4.11. Just a couple questions/comments:

* With meson, will we still need to use the setup.py at all? There is one non-standard thing that we did by building the python doc strings [here](https://github.com/SHTOOLS/SHTOOLS/pull/380/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L107-L111) and this will probably need to be integrated into the meson workflow somehow.
* Will pyshtools be installed using pip, as before, or do we need to add the `meson compile` command somewhere (such as in the github workflows)?
* And last, is there anything in this PR that we should include with v4.10.2?

pyshtools/meson.build
QuLogic: I guess this requires running with `--no-build-isolation`, because I don't have this problem by default.

QuLogic:
> * With meson, will we still need to use the setup.py at all? There is one non-standard thing that we did by building the python doc strings [here](https://github.com/SHTOOLS/SHTOOLS/pull/380/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L107-L111) and this will probably need to be integrated into the meson workflow somehow.

No, I will delete `setup.py` once I have confirmed that the produced wheels are equivalent. One should not be using it directly these days anyway; tools like `pip` or `build` that implement the newer PEPs are the way to do installs/builds now.
 
> * Will pyshtools be installed using pip, as before, or do we need to add the `meson compile` command somewhere (such as in the github workflows)?

Yes, regular Python installs will still use wheels that can be installed by `pip`. For building from source (tarball/git), you can also use `pip`, but for generating wheels only it's probably better to use `build`.

pyshtools/meson.build
QuLogic: Yes, I only have in-tree venvs.

QuLogic:
This is now fully `pip install`-able, and I can run the examples from that venv. An editable install also works if using `--no-build-isolation --editable` as documented by meson-python. The docs now work as well, though not in editable mode. The only thing left to do is figure out what to do with that pre-generated `_version.py`.

QuLogic:
> The only thing left to do is figure out what to do with that pre-generated `_version.py`.

There seem to be a few options for this, though it's not entirely builtin to meson-python: https://github.com/mesonbuild/meson-python/issues/159; e.g., [this script in Pandas](https://github.com/pandas-dev/pandas/blob/main/generate_version.py) or [SciPy using `.dev0` by default but git version at runtime](https://github.com/scipy/scipy/blob/main/meson.build#L7).

Also, I just realized that the docs are not compatible with cross-compilation, as it requires importing the shared library. But that's something that already existed and I guess you just don't support that.

MarkWieczorek:
Could you rebase this off of the latest v4.10.3 release that went out a couple of days ago? That will be the final release using distutils.

In order to get the last release to work with macos arm64 and conda cross compiling, I needed to make a couple changes that will conflict here:

1. We will need to add the folder `pyshtools/doc` to the meson build. This includes the python docstrings that are now included directly in the repo (as opposed to being generated at build time). As part of this, `make_docs.py` was moved to the `doc` folder.
2. The make_docs part of the main meson.build file will probably need to be removed because of (1).
3. Some changes were made to the setup.py file, but I think that we are just going to delete that anyways.
4. For the `_version` file I recently ran into a problem with versioneer in a different project of mine, and there I just switched to `setuptools_scm`. I don't know if this will solve the problems here, but it was trivial to change: we just need to remove the versioneer.py file, the `pyshtools/_version.py` file, the config part in the `pyproject.toml` file, and then add this to the toml file:
```
[tool.setuptools_scm]
write_to = "pyshtools/_version.py"
version_scheme =  "post-release"
local_scheme =  "no-local-version"
```
and then add this to the pyshtools `__init__` file:
```
from importlib.metadata import version, PackageNotFoundError

try:
    __version__ = version('ctplanet')
except PackageNotFoundError:
    # package is not installed
    pass
```

I'll go through this next week. It shouldn't take too much time now that I kind of understand how meson works.

FlorianPfaff:
Hi everyone,

just in case it is of help to anyone here. I had problems compiling the meson branch but was able to get it working. This concerns two points:
1. Building a wheel would always fail with an error.
2. lapack-compatible openblas installations are not found.


# 1. Building a wheel fails with an error
Building a wheel failed for me on both Ubuntu 20.04 and Alpine Linux with the same error message. Here goes the log:

```text
     The Meson build system
      Version: 1.1.1
      Source dir: /home/runner/work/pyRecEst/pyRecEst/SHTOOLS
      Build dir: /home/runner/work/pyRecEst/pyRecEst/SHTOOLS/.mesonpy-nrqahxw_/build
      Build type: native build
      Project name: shtools
      Project version: 4.11.0
      Fortran compiler for the host machine: gfortran (gcc 12.2.1 "GNU Fortran (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924")
      Fortran linker for the host machine: gfortran ld.bfd 2.40
      C compiler for the host machine: cc (gcc 12.2.1 "cc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924")
      C linker for the host machine: cc ld.bfd 2.40
      Host machine cpu family: x86_64
      Host machine cpu: x86_64
      Found pkg-config: /usr/bin/pkg-config (1.9.5)
      Found CMake: /usr/bin/cmake (3.26.4)
      Run-time dependency openblas found: NO (tried pkgconfig and cmake)
      Run-time dependency blas found: YES
      Run-time dependency lapack found: YES 3.11.0
      Run-time dependency fftw3 found: YES 3.3.10
      Program python found: YES (/usr/bin/python3)
      Program make_docs.py found: YES (/home/runner/work/pyRecEst/pyRecEst/SHTOOLS/pyshtools/doc/make_docs.py)
      C++ compiler for the host machine: c++ (gcc 12.2.1 "c++ (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924")
      C++ linker for the host machine: c++ ld.bfd 2.40
      Build targets in project: 17
      
      shtools 4.11.0
      
        User defined options
          Native files   : /home/runner/work/pyRecEst/pyRecEst/SHTOOLS/.mesonpy-nrqahxw_/build/meson-python-native-file.ini
          buildtype      : release
          default_library: static
          b_ndebug       : if-release
          b_vscrt        : md
      
      Found ninja-1.11.1.git.kitware.jobserver-1 at /tmp/pip-build-env-p4f3n37j/normal/bin/ninja
      Traceback (most recent call last):
        File "/usr/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/usr/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 152, in prepare_metadata_for_build_wheel
          whl_basename = backend.build_wheel(metadata_directory, config_settings)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-p4f3n37j/overlay/lib/python3.11/site-packages/mesonpy/__init__.py", line 1048, in wrapper
          return func(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-p4f3n37j/overlay/lib/python3.11/site-packages/mesonpy/__init__.py", line 1113, in build_wheel
          return project.wheel(out).name
                 ^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-p4f3n37j/overlay/lib/python3.11/site-packages/mesonpy/__init__.py", line 969, in wheel
          file = self._wheel_builder.build(directory)
                 ^^^^^^^^^^^^^^^^^^^
        File "/usr/lib/python3.11/functools.py", line 1001, in __get__
          val = self.func(instance)
                ^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-p4f3n37j/overlay/lib/python3.11/site-packages/mesonpy/__init__.py", line 792, in _wheel_builder
          self._install_plan,
          ^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-p4f3n37j/overlay/lib/python3.11/site-packages/mesonpy/__init__.py", line 860, in _install_plan
          if details['tag'].strip() not in install_tags:
             ^^^^^^^^^^^^^^^^^^^^
      AttributeError: 'NoneType' object has no attribute 'strip'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```


After removing the [tool.meson-python.args] section, python -m build works just fine.


# 2. lapack-compatible openblas installations are not found.

This is pretty annoying when, e.g., using Alpine linux, where the lapack-compatible openblas package conflicts with the regular lapack package.
Since meson-python uses pkg-config and cmake to find the dependencies, I appended LAPACK to the list of dependencies that fullfil the lapack requirement. cmake finds the lapack-compatible openblas package when looking for "LAPACK", it does *not* find it when looking for "lapack" (capitalization matters here).

I have created a PR to the meson branch of the fork that this PR is about. I hope this is of some help and/or prevents some frustration for someone else.

# Reminders

- [X] Base all changes on the `develop` branch: the `master` branch is used only when releasing new versions.
I based it on the meson branch, which should perfectly make sense
- [X] Run `make check` to ensure that the python code follows standard formatting conventions.
No Python code was changed.
- [X] If adding new features, update the docstring to provide all information that is required to use the feature.
No new features

rgommers:
> 1. Building a wheel would always fail with an error.

It looks like this is another occurrence of https://github.com/mesonbuild/meson-python/issues/421, which was reported a few days ago. That needs to be fixed in the next meson-python release.

> 2. lapack-compatible openblas installations are not found.

Auto-detecting one of multiple possible options for BLAS or LAPACK libraries in Meson is still a work in progress, see https://github.com/mesonbuild/meson/pull/10921. I'm hoping to wrap that up within the next two months. That should also take care of Meson and CMake dealing with capitalization differently. For now, you need a CLI build option, and handling it like what SciPy does: https://github.com/scipy/scipy/blob/main/scipy/meson.build#L151-L179.

FlorianPfaff:
> For now, you need a CLI build option, and handling it like what SciPy does: https://github.com/scipy/scipy/blob/main/scipy/meson.build#L151-L179.
Hi,

Thank you for the informative reply. I was already wondering if it might be an issue of meson-python, but was thinking it may also be likely to be an issue of SHTOOLS since you are still in the process of switching to this build system. No worries, I made a [fork and found a working solution for it](https://github.com/FlorianPfaff/SHTOOLS/tree/meson). Since I did not know it was an issue of meson-python, I [created a PR ](https://github.com/QuLogic/SHTOOLS/pull/1) for that branch, so you could take a look what I had to change to make it work (maybe you could also consider changes to SHTOOLS to complete your transition to meson before meson-python makes the fixes) - but that is of course up to you.
Cheers,

Florian

QuLogic:
I've rebased, and fixed some issues:

> * We will need to add the folder `pyshtools/doc` to the meson build. This includes the python docstrings that are now included directly in the repo (as opposed to being generated at build time). As part of this, `make_docs.py` was moved to the `doc` folder.
> 
> * The make_docs part of the main meson.build file will probably need to be removed because of (1).

I've removed this from the `meson.build` file; I assume you will continue to refresh those files using the existing `make doc`. I've modified `make_docs.py` so that it refreshes the `meson.build` file with the list of documentation files as well.

> * Some changes were made to the setup.py file, but I think that we are just going to delete that anyways.

Yes, I think that's all unneeded.

> * For the `_version` file I recently ran into a problem with versioneer in a different project of mine, and there I just switched to `setuptools_scm`. I don't know if this will solve the problems here, but it was trivial to change: we just need to remove the versioneer.py file, the `pyshtools/_version.py` file, the config part in the `pyproject.toml` file, and then add this to the toml file:

I haven't done this bit yet, but can. I've noticed that versioneer's generated file is not fully working because meson-python does not use Meson's dist tarball https://github.com/mesonbuild/meson-python/pull/58 and breaks `export-subst`. I will have to check if `setuptools_scm` uses the same mechanism or not.

> > 1. Building a wheel would always fail with an error.
> 
> It looks like this is another occurrence of [mesonbuild/meson-python#421](https://github.com/mesonbuild/meson-python/issues/421), which was reported a few days ago. That needs to be fixed in the next meson-python release.

While there was a bug in meson-python, I think this was actually my fault as I forgot to put an install tag on the docs.

> > 2. lapack-compatible openblas installations are not found.
> 
> Auto-detecting one of multiple possible options for BLAS or LAPACK libraries in Meson is still a work in progress, see [mesonbuild/meson#10921](https://github.com/mesonbuild/meson/pull/10921). I'm hoping to wrap that up within the next two months. That should also take care of Meson and CMake dealing with capitalization differently. For now, you need a CLI build option, and handling it like what SciPy does: https://github.com/scipy/scipy/blob/main/scipy/meson.build#L151-L179.

There is a build option already.

MarkWieczorek:
Could you take a look at the new CI runs, that are now passing? I fixed a small problem in the CI files where mamba was throwing an error for no reason.

My plan is to put out a new minor release next week (4.10.4) that fixes a few bugs, then include this PR in the next major release (4.11) after I do some testing...

environment-dev.yml
```diff
@@ -1,8 +1,9 @@
 channels:
     - conda-forge
 dependencies:
+    - pkg-config
     - fftw>=3.3.8
```
MarkWieczorek, line 5: Shouldn't we also add meson-python as a dependency as its need to build from source?

environment.yml
```diff
@@ -1,8 +1,9 @@
 channels:
     - conda-forge
 dependencies:
+    - pkg-config
     - fftw>=3.3.8
```
MarkWieczorek, line 5: Shouldn't we also add meson-python as a dependency as its need to build from source?

examples/fortran/SHCilmPlus/meson.build
```diff
@@ -0,0 +1,11 @@
...
+  'TestCilmPlus',
+  'TestCilmPlus.f95',
+  link_with: shtools)
+
+test(
+  'CilmPlus',
+  TestCilmPlus,
+  args: [
+    meson.current_source_dir() / '../../ExampleDataFiles',
+  ])
```
MarkWieczorek, line 11: This test timed out on my laptop. I would suggest adding this argument to `test`:
```
timeout: 60
```

examples/fortran/MarsCrustalThickness/meson.build
```diff
@@ -0,0 +1,12 @@
...
+  'MarsCrustalThickness.f95',
+  link_with: shtools)
+
+test(
+  'MarsCrustalThickness',
+  TestMarsCrustalThickness,
+  args: [
+    meson.current_source_dir() / '../../ExampleDataFiles',
+    meson.current_source_dir() / 'input.txt',
+  ])
```
MarkWieczorek, line 12: At this point only for the fortran programs are being tested, but for the pyshtools distribution we should probably test the python components. Similar to what is done in the Makefile, I think that we should probably create a few test "suites" for: fortran, fortran-mp, fortran-no-timing, python, python-no-timing. These would be specified by adding the keyword `suite: ['fortran', 'fortran-no-timing']` to `test`.

I will probably just make separate PR on your branch in the next day or two (unless you get to this first!)

examples/fortran/SHLocalizedAdmitCorr/meson.build
```diff
@@ -0,0 +1,12 @@
...
+  'LocalizedAdmitCorr.f95',
+  link_with: shtools)
+
+test(
+  'LocalizedAdmitCorr',
+  TestLocalizedAdmitCorr,
+  args: [
+    meson.current_source_dir() / '../../ExampleDataFiles',
+    meson.current_source_dir() / 'input.txt',
+  ])
```
MarkWieczorek, line 12: This test timed out on my laptop. I would suggest adding this argument to `test`:
```
timeout: 60
```

examples/fortran/SHMag/meson.build
```diff
@@ -0,0 +1,11 @@
...
+  'SHMag',
+  'SHMag.f95',
+  link_with: shtools)
+
+test(
+  'SHMag',
+  TestSHMag,
+  args: [
+    meson.current_source_dir() / '../../ExampleDataFiles',
+  ])
```
MarkWieczorek, line 11: This test timed out on my laptop. I would suggest adding this argument to `test`:
```
timeout: 60
```

examples/fortran/TimingAccuracy/meson.build
```diff
@@ -0,0 +1,30 @@
...
+}
+foreach name, args : BENCHS
+  test_name = args[0]
+  input = args[1]
+  benchmark(
+    name,
+    timing_tests[test_name],
+    args: [
+      meson.current_source_dir() / input,
+    ])
```
MarkWieczorek, line 29: These benchmarks will certaintly timeout, as the default is 30 seconds. I suggest adding no time limit by adding
```
timeout: 0
```

examples/meson.build
```diff
@@ -0,0 +1,2 @@
+subdir('cpp')
+subdir('fortran')
```
MarkWieczorek, line 2: Note: we should also add the "notebooks" and "python" directories as well, after we add the meson.build files to these.

meson_options.txt
```diff
@@ -0,0 +1,4 @@
+option('blas', type: 'string', value: '',
```
MarkWieczorek, line 1: According to https://mesonbuild.com/Build-options.html#build-options, this file should be renamed "meson.options" for meson version 1.1+. (Both seem to work, but the old name could get deprecated at some point). If we use meson.options, then we should need to ensure that the proper version of meson-python is being used, which is 0.14.0+

.github/workflows/pypi_publish.yml
```diff
@@ -31,12 +31,11 @@ jobs:
       shell: bash -l {0}
       run: |
         mamba config --set always_yes yes --set changeps1 no
-        mamba install -q -n pypi fftw \
-                                 liblapack \
-                                 openblas \
+        mamba install -q -n pypi blas-devel \
+                                 fftw \
+                                 meson-python \
```
MarkWieczorek, line 36: should we require meson-python to be >= 0.14.0 ? As noted below, the convention of using the file name `meson.options` instead of `meson_options.txt` is for meson 1.1.+ and meson-python 0.14.+

examples/fortran/TimingAccuracy/meson.build
MarkWieczorek: Also, the TimingAccuracy source files were never modified to accept runtime arguments. If you run `meson test --benchmark`, it will just wait forever to recieve user input from stdin.

I will probably make a separate PR on your branch for this in the next day or two.

meson.build
```diff
@@ -0,0 +1,25 @@
+project('shtools', 'c', 'fortran',
+  version: '4.11.0dev0',
+  license: 'BSD-3-Clause',
+  meson_version: '>=1.0.0')
```
MarkWieczorek, line 4: I think that we should use version 1.1.0+ in order to be compatible with the file `meson.options`.

meson.build
```diff
@@ -0,0 +1,25 @@
+project('shtools', 'c', 'fortran',
+  version: '4.11.0dev0',
+  license: 'BSD-3-Clause',
```
MarkWieczorek, line 3: The meson and meson-python docs are a little confusing and inconsistent as to what the default buildtype and optimization level is. meson claims that the default buildtype is debug optimize and meson-python claims it is release. meson-python claims that the default optimization is 2, but when I checked `meson configure`, it said that we were using level 3. It might be prudent to set the following here:

```
default_options: [
  'buildtype=release',
  'optimization=3',
]
```

pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+  'Topic :: Scientific/Engineering',
+  'Topic :: Scientific/Engineering :: GIS',
+  'Topic :: Scientific/Engineering :: Mathematics',
+  'Topic :: Scientific/Engineering :: Physics',
+]
+authors = [
+  { name = 'The SHTOOLS developers', email = 'mark.wieczorek@ipgp.fr' },
+]
+dependencies = [
+  'numpy',
```
MarkWieczorek, line 32: is meson-python a dependency here?

pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+[project.optional-dependencies]
+cartopy = ['cython', 'pyshp', 'six', 'shapely', 'cartopy>=0.18.0']
+pygmt = ['pygmt>=0.3']
+palettable = ['palettable>=3.3']
+ducc = ['ducc0>=0.15']
+
+[build-system]
+build-backend = 'mesonpy'
+requires = [
+  'meson-python>=0.13.1',
```
MarkWieczorek, line 55: I would probably change this to 0.14.0+

pyproject.toml
MarkWieczorek: Also, I am not 100% sure if we should add the following or not:

pkg-config: required by meson-python I think.
jupyter: necessary to run the notebooks, but nothing else

blas-devel
fftw

these two could be installed using a system package manager, but maybe we want to force pyshtools to use these packages anyways?

pyshtools/_version.py
```diff
@@ -5,8 +5,9 @@
 # directories (produced by setup.py build) will contain a much shorter file
```
MarkWieczorek, line 5: I don't think that versioneer works properly without a setup.py file. I will look into using setuptools scm instead.

pyshtools/meson.build
```diff
@@ -0,0 +1,76 @@
+python_sources = [
+  '__init__.py',
+  '_version.py',
```
MarkWieczorek, line 3: we may need to remove this if switching from versioneer to scm

pyshtools/meson.build
```diff
@@ -0,0 +1,76 @@
...
+
+incdir_f2py = incdir_numpy / '..' / '..' / 'f2py' / 'src'
+inc_f2py = include_directories(incdir_f2py)
+fortranobject_c = incdir_f2py / 'fortranobject.c'
+
+
+pyshtools_pyf = custom_target('pyshtools_c',
+  input: '../src/pyshtools.pyf',
+  output: ['_SHTOOLS-f2pywrappers.f', '_SHTOOLSmodule.c'],
+  command: [python, '-m', 'numpy.f2py', '@INPUT@', '--build-dir', '@OUTDIR@'])
```
MarkWieczorek, line 52: I don't think that OUTDIR is defined anywhere, but maybe this is ok?

MarkWieczorek (commented):
This looks great, and I got it to compile on Fedora with no real problems. Besides the comments I made on the source code, I think that these are the only things left to do:

1. Fix the fortran TimingAccuracy source file to accept command line arguments. I will do this next and open a PR on your branch.
2. Add options to run the python tests and python notebooks. This will likely be done by using the "suite" keywords in the test() functions. I'll do that after I fix the fortran tests and then make a PR on your branch.
3. Get the cpp test to build. (Not sure what is going on there...)
4. Replace the versioneer system with setuptools scm (or something else). Also try to set the version dynamically in the main meson.build file (not sure if that is possible though).

After this, we can put out a new release and see if the conda packages get built...

pyproject.toml
eli-schwartz: pyproject.toml "dependencies" is exclusively for runtime pip-installable python library dependencies.

Meson-python is a build time dependency, not a runtime one.

Pkg-config (and development headers) are build time dependencies but also not python dependencies: they must be installed by the system (an OS or conda package manager).

There's an in-progress PEP to define a brand new way to specify external system dependencies: https://peps.python.org/pep-0725/

This, if/when accepted, will be the right place to list things like BLAS headers.

pyshtools/meson.build
eli-schwartz: This may look like something similar to a python f-string, but it is not.

It's an ordinary string passed to custom_target's "command" parameter, which in turn accepts an array of strings which will be processed as templates. OUTDIR is one of the documented template replacements.

meson.build
eli-schwartz: This is because meson-python will by default act in the same position as an end user, and pass configurations to meson to use what meson considers a non-default build type.

MarkWieczorek:
@eli-schwartz Could you take a quick look at this PR that changes pyshtools from using versioneer to setuptools_scm? https://github.com/QuLogic/SHTOOLS/pull/4 

This works when run from a git repo, but I have some doubts as it doesn't work with source tarballs. If this isn't good, I will try to get things to work with versioneer instead...

MarkWieczorek:
I just finished 4 small PRs on @QuLogic's meson branch:
* https://github.com/QuLogic/SHTOOLS/pull/2
* https://github.com/QuLogic/SHTOOLS/pull/3
* https://github.com/QuLogic/SHTOOLS/pull/4
* https://github.com/QuLogic/SHTOOLS/pull/5

I think that this is probably all that needs to be done for a next release with python 3.12. Any comments would be appreciated.

MarkWieczorek:
FYI, I'm going to work on putting out a new release with the meson build backend on Jan. 15-16. If you have anything to add, please comment here or on the merged branch here : https://github.com/SHTOOLS/SHTOOLS/pull/415

PR #391, opened 2023-05-09
Use raw strings for regex

QuLogic:
This prevents several invalid escape warnings, and though it isn't necessary if there aren't any backslashes in the string, it's easier to be consistent about it.

**Reminders**


MarkWieczorek:
Could you rebase this off of the v4.10.3 release that just went out? I think that the only conflict is that `pyshtools/make_docs.py` is now located at `pyshtools/doc/make_docs.py`.

PR #392, opened 2023-05-12
prebuild python docstrings and store in repo

MarkWieczorek:
This PR stores the python docstrings for wrapped functions in the repo, as opposed to creating them at build time. This should help to get the macOS ARM conda builds to work, and is also necessary for our transition from distutils to meson. 

This is intended to be the last shtools version that uses `setup.py` and distutils. All further releases will use meson.


MarkWieczorek:
Solved by adding 
```
package_data={'pyshtools': ['doc/*.doc']},
```
to `setup.py` metadata.

PR #397, opened 2023-08-08
Fix an issue with the southern hemisphere calculations in MakeMagGridDH and MakeGravGridDH

AB-Ares:
Both MakeMagGridDH and MakeGravGridDH return incorrect tensors for the southern hemisphere. 

To demonstrate this, I used the pyshtools SHTensor, which uses MakeGravGridDH, class to compute the Bouguer tensor of the lunar gravity field at the north pole. In the figure below, I show a north polar orthographic projection of the 5 components of the tensor in the upper panels. In the middle panels, I first rotated the SHGravCoeffs to the south pole, computed the tensor there, and rotated it back to the north pole. In principle, the upper and middle rows should give the exact same results, but some artifacts are clearly seen when the tensor is computed in the southern hemisphere, as shown in the difference plots of the lower panels. These artifacts were found for all planetary bodies, for different tested grid resolutions, and also for the Magnetic field tensor.

Grav_NorthPole_Moon_notfixed

Looking at the fortran source code for MakeGravGridDH, I have noticed 2 typos:
– One in the symmetry of the 2nd order derivative of the Legendre Polynomials (dpl2s)
– And in the computation of the south hemisphere tensor components, where the northern hemisphere colatitude is used instead of that of the south. 

With these fixes, the procedure described above returns the same exact tensor, as shown below.
Grav_NorthPole_Moon_fix

The same fixes have been applied to MakeMagGridDH, and corrected the artifacts in the Mag tensor. 

**Reminders**


MarkWieczorek:
Note that the CI tests are failing because of a bug with mamba: https://stackoverflow.com/questions/76847652/mamba-env-throws-attributeerror-namespace-object-has-no-attribute-func

Issue #401, opened 2023-08-09
NameError: name 'shcoeffs' is not defined


EricAtORS:
**Description of the problem**
I have a circular import problem on windows with python 3.10.12 and pyshtools 4.10.3.
I don't know why it works else where, but not for me.

You can see that it's trying to do a circular import.
First, when you import `pyshtools.datasets` in the init. Then, `datasets.Merceury` tries to import `SHCoeffs ` from `pyshtools.shclasses`.
But, then `pyshtools.shclasses` can only be imported after `pyshtools.datasets`.

I tried swapping the order of the 2 and it worked for me.

```
In [3]: from pyshtools.shclasses import shcoeffs
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 from pyshtools.shclasses import shcoeffs

File C:\Perforce\Development\Dragonfly\dragonfly_python_environment_windows\lib\site-packages\pyshtools\__init__.py:48
     46 from . import backends
     47 from . import constants
---> 48 from . import datasets
     49 from . import shclasses
     50 from . import legendre

File C:\Perforce\Development\Dragonfly\dragonfly_python_environment_windows\lib\site-packages\pyshtools\datasets\__init__.py:19
      1 """
      2 pyshtools datasets.
      3
   (...)
     17 use of the variable nt.
     18 """
---> 19 from . import Mercury
     20 from . import Venus
     21 from . import Earth

File C:\Perforce\Development\Dragonfly\dragonfly_python_environment_windows\lib\site-packages\pyshtools\datasets\Mercury.py:19
     17 from pooch import retrieve as _retrieve
     18 from pooch import HTTPDownloader as _HTTPDownloader
---> 19 from ..shclasses import SHCoeffs as _SHCoeffs
     20 from ..shclasses import SHGravCoeffs as _SHGravCoeffs
     21 from ..constants.Mercury import omega as _omega

File C:\Perforce\Development\Dragonfly\dragonfly_python_environment_windows\lib\site-packages\pyshtools\shclasses\__init__.py:85
     81 from .shtensor import SHMagTensor
     83 from .shgradient import SHGradient
---> 85 del shcoeffs  # noqa: F821
     86 del shgrid  # noqa: F821
     87 del shwindow  # noqa: F821

NameError: name 'shcoeffs' is not defined
```

**System information**

PR #402, opened 2023-08-09
Break circular imports. #401

EricAtORS:
This breaks the circular import reported in #401 

**Reminders**

### DIFF
--- a/pyshtools/__init__.py
+++ b/pyshtools/__init__.py
@@ -45,8 +45,8 @@ from ._version import get_versions as _get_versions
 # ---- Import shtools subpackages ----
 from . import backends
 from . import constants
-from . import datasets
 from . import shclasses
+from . import datasets
 from . import legendre
 from . import expand
 from . import shio


MarkWieczorek:
Thanks for this. I don't understand why this works everywhere else, but if if solves your problem, great!

I'm not sure why the CI builds are failing. I suspect it has something to do with mamba (its certainly not your change...)

EricAtORS:
@MarkWieczorek Thanks for merging it in so quickly.
Is there anything that I can do to help push that time line up? Mostly just looking for that bug fix.

MarkWieczorek:
I need to go through the big PR about our upcoming transition to using the Meson build system, as this might be incorporated in the next release if it all works. I also would like to track down the bug that is crashing the CI tests (see https://github.com/SHTOOLS/SHTOOLS/actions/runs/5813995031/job/15777794561). If you know anything about either of those, that would help.

Putting out new releases is always a pain: half the time it goes smoothly, the other half I need to spend a lot of time updating build procedure. The windows builds are the most problematic, though the last two times they were pretty easy. I just need to find a week where I have the time to invest in this in case something goes wrong. The best time would probably be the end of august...

EricAtORS:
Perfect.

For the build issue,
I'm not familiar with github actions, but it's definitely something I'm looking in to learning.

But I've always seen similar issues when there's a desync between mamba and conda. Either mamba needs to be updated or conda needs to be held back if mamba hasn't updated with the newest change.

I'll see what I can do with this.

For windows, feel free to tag me.
I was also the one who pushed the changes on conda-forge.
I was also trying to package it up in conan for our internal use, but managing fortran on windows was not that straight forward... so I can feel your pain.

PR #439, opened 2024-01-23
New method for setting the version using versioneer in meson.build

MarkWieczorek:
This PR removes the `setuptools_scm` approach of obtaining the version number and replaces it with a slightly modified approach using `versioneer`.

The problem with setuptools_scm is that this line in meson.build returns an error when run from within a non git versioned directory (such as generated by a source tarball):
```
version: run_command(find_program('python3'), '-m', 'setuptools_scm').stdout().strip(),
```

The modified approach uses a call to versioneer instead
```
version: run_command(find_program('python3'), '-c', 'import os; os.chdir("pyshtools"); import _version; print(_version.get_versions()["version"])').stdout().strip(),
```
As part of this change, we needed to (re-)add the `pyshtools/_version.py` file that was found in v4.10.x.

There are two potential problems with this approach. 

The first is that versioneer really doesn't support the meson-python backend, and when you create the non-vendored `_version.py` file, you get an error
```
$ versioneer install --no-vendor
 creating pyshtools/_version.py
 pyshtools/__init__.py unmodified
Traceback (most recent call last):
  File "/Users/lunokhod/miniconda3/envs/py312/bin/versioneer", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/lunokhod/miniconda3/envs/py312/lib/python3.12/site-packages/versioneer.py", line 2001, in main
    setup_command()
  File "<string>", line 2270, in setup_command
  File "<string>", line 2235, in scan_setup_py
FileNotFoundError: [Errno 2] No such file or directory: 'setup.py'
```
The error is not important. However, it also adds the following to the `__init__.py` file
```
from . import _version
__version__ = _version.get_versions()['version']
```
which I removed and instead kept the existing 
```
try:
    __version__ = version('pyshtools')
except PackageNotFoundError:
    # package is not installed
    pass
```
The second problem is that when building from within a source tarball generated by a git archive command, meson set the version to `Successfully installed pyshtools-0+unknown`. This is better than a failure, but there is very little additional information. 

If you run 
```
python -c 'import os; os.chdir("pyshtools"); import _version; print(_version.get_versions())'
{'version': '0+unknown', 'full-revisionid': '9ec26db51f2f7a836b4c346e7f16328884a870f9', 'dirty': False, 'error': 'no suitable tags', 'date': None}
```
you will see that it actually gets the commit, which is useful. However, it doesn't get the date, even though the date does get hardcoded in `_version.py` file when creating the source tarball.

```
    git_refnames = " (new_versioneer)"
    git_full = "9ec26db51f2f7a836b4c346e7f16328884a870f9"
    git_date = "2024-01-23 12:54:16 +0100"
    keywords = {"refnames": git_refnames, "full": git_full, "date": git_date}
```
Also, even though I haven't dug too deep in the `_version.py` file, I was somewhat surprised that it didn't deduce the version number from the name of the source directory (this is one of the methods they use to get a version number).

If we keep this approach, I will probably add something like this to the `__init__.py` file
```
__commit__ = _get_versions()["full-revisionid"]
``` 
so that the user can get at least some info after importing the package. However, I would really like versioneer to produce something like `v4.11.10.devN`.


MarkWieczorek:
I'm going to close this, and the modifications to `setuptools_scm` in this PR resolves these issues: https://github.com/SHTOOLS/SHTOOLS/pull/460

PR #441, opened 2024-01-30
Develop

MarkWieczorek:
* Changed the behavior of the fortran `MakeCircleCoord` routine when the cirlce radius was 0. Before, this would output a single set of coordinates. However, in pyshtools, an array would be returned of length 360/interval with all but the first being (0,0). Now, an array of coordinates of length 360/interval is returned where all coordinates are the same. 
* Added `jupyter` to the pyproject.toml build-system.requirements (it is required for the tests, even if the tests aren't run by default).

**Reminders**

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ['3.8', '3.12']
+        python-version: ['3.9', '3.12']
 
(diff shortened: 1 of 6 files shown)


QuLogic:
> Added `jupyter` to the pyproject.toml build-system.requirements (it is required for the tests, even if the tests aren't run by default).

BTW, for the `jupyter` executable, I _think_ you only need `jupyter-core`, which trims the dependency tree slightly. But I did not check what all the other dependencies of the full package are, and if they'd be needed.

PR #448, opened 2024-02-27
Remove Cartopy build-time dependencies

QuLogic:
Latest Cartopy produces wheels, so there's no need to install its build-time dependencies ourselves.

**Reminders**


MarkWieczorek:
Does Cartopy need proj and geos? We should update the "how to install cartopy" section here: https://github.com/QuLogic/SHTOOLS/blob/no-cartopy-deps/docs/pages/mydoc/python-installing.md

If this is all moot, we can just delete the whole section.

QuLogic:
That's all hidden behind `pyproj`, which has a pre-built wheel. As [shown in the Cartopy link there](https://scitools.org.uk/cartopy/docs/latest/installing.html#required-dependencies), no proj or geos any more.

MarkWieczorek:
looks good. Thanks!

PR #449, opened 2024-03-02
Add pathlib support to input and output

jschmaia:
With this change you can pass `pathlib.Path` to `shread` and other input/output functions, including `SHCoeffs.to/from_file`, `SHGravCoeffs.to/from_file` and `SHMagCoeffs.to/from_file`.

### DIFF
--- a/pyshtools/shclasses/shcoeffs.py
+++ b/pyshtools/shclasses/shcoeffs.py
@@ -1046,7 +1046,7 @@ class SHCoeffs(object):
 
         if isinstance(filename, Path):
             filename = str(filename)
-        
+
         if filename[-3:] == '.gz':
             filebase = filename[:-3]
         else:
--- a/pyshtools/shclasses/shcoeffs.py
+++ b/pyshtools/shclasses/shcoeffs.py
@@ -12,6 +12,7 @@ import shutil as _shutil
 import warnings as _warnings
 from scipy.special import factorial as _factorial
 import xarray as _xr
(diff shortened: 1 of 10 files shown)


MarkWieczorek:
This looks good to me! But could you also make a few small changes so that this works in other routines as well?

* `shclasses.shgeoid.to_netcdf()` - we just need to update the docstring so that people know you can also use a Path.
* `shio.shwrite()`, `read_igrf()`, `read_icgem()`, `read_dov()`, `write_dov()`, `read bshc()`, `write_bshc()` - just copy to path to string conversion at the begging of each method.

And...

I noticed the`shcoeffs.to_file()` is a bit odd when using `.gz` files. This calls the shio method, which can deal with .gz files fine. However, `shcoeffs.to_file()` writes a text file by calling `shio.write_xxx` and then gzips it afterwards. I think that we can just remove all the gz and filebase stuff in `shcoeffs.to_file()` and just pass the Path directly to the dedicated write routine. If you don't want to make so many changes, I can easily do this myself later.

jschmaia:
I did a new commit with the small changes you suggested:

- updating the docstring of `shclasses.shgeoid.to_netcdf()`
- adding the path to str conversion and associated docstring changes to the methods: `read_bshc()`, `write_bshc()`, `read_dov`, `write_dov`, `read_icgem_gfc()`, `write_icgem_gfc()`, `read_igrf()`,`shio.shwrite()`

PR #458, opened 2024-03-18
Fix tensor to xarray

spicydonkey:
This PR fixes `Tensor.to_xarray` by using the correct unit attribute.

- Replaces the undefined attribute `self._vii_units` in `Tensor` by `self.units`.
- Adds a demo using the fixed `Tensor.to_xarray` method in the tutorial notebook `gravity-and-magnetic-fields.ipynb`, with a brief description of xarray.

**Reminders**


spicydonkey:
Bumping this one please @MarkWieczorek .

MarkWieczorek:
Thanks for fixing this, and sorry about the delay. We've been busy with some packaging issues with the latest release.

I'm ok with the changes regarding the units, but could we make a couple minor changes to the jupyter notebook? I'm mostly ok with this, but I think that we only need to demonstrate and print the `to_xarray` method once, and just note that this also works with the geoid and tensors as well. 

For cell 8, I just have a couple minor tweaks

```
Most `pyshtools` classes that store grids have a method `to_xarray` that returns an `xarray.Dataset` containing the coordinates, one or more variables (such as the components of the gravity vector), and metadata. [`xarray`](https://docs.xarray.dev/en/stable/) is a powerful tool for working with multi-dimensional arrays, and it is particularly useful for working with geophysical data, which are conventionally written to file in NetCDF format. 

Here, we use the `to_xarray` method to create a `Dataset` and print its contents:
```
(this is just a sujestion)

Then I would probably put "Remember, you should never use the 'rainbow' and 'jet' colormaps! Colormaps are discussed more in the notebook 'Plotting Maps'." just before cell 8 and after the image that demonstrates how to change the colormap.

For the geoid, it's probably ok to show how to create the xarray, but I don't think that we need to print it, as it doesn't really give any more info than when we printed the gravity xarray. I would probably remove the print statement for tensor as well.

If you think its useful to print the geoid and tensor xarray, maybe it would be *more* useful to demonstrate another feature of xarray. For example, we could print the coordinates, or maybe save an individual variable to file (though we can also do that with `to_netcdf` as well.) Or maybe just some simple mathematical operator (like add/subtract GM/r2 from a grid. Or whatever peope use xarray for (I don't use it that often!).

spicydonkey:
Thanks for the review @MarkWieczorek , and keeping up your work in shtools! 

The cell 8 has been updated with your suggestion.

The tip on colormap has been moved to the location you suggested, and blends nicer in the tutorial.

I agree with your idea that there isn't an added benefit to the tutorial of demonstrating `to_xarray` more than once.  I like using xarray since it provides a dataframe like object.  And as a novice in geodata, the familiar data structure helps in the exploratory phase.  I do look foward to using shtools' built-in features.

PR #469, opened 2024-04-19
Migration to cibuildwheel

MarkWieczorek:
This PR is part of our efforts to migrate away from multibuild to cibuildwheel in order to build our wheels for distribution on pypi.org

Currently, the following are working:
* macos (intel and ARM)
* linux (manylinux)

The following are not working
* windows
* aarch64 (linux ARM)

The gcc, openblas and fftw dependencies are being installed by the system package managers (brew for macos, and yum for CentOS). I have no idea how to do this for windows, and am unlikely to invest the time in trying. Help would be appreciated!

For aarch64 (which should work on asahi linux running on apple hardware), this is built by cross compilation. Thus, installing the dependencies on the host by the system package manager probably won't work. It's possible that we might need to install fftw and openblas from source. If anyone has experience with this, please let me know the best way to do this.

**Reminders**


MarkWieczorek:
Closed, as this is incorporated into https://github.com/SHTOOLS/SHTOOLS/pull/479

PR #479, opened 2024-06-30
Migration to cibuildwheel (Windows & Linux aarch64 build support)

cometbeetle:
This includes all the changes of #469, but with Windows build support. 

It also disables builds for PyPy Python (meaning we only generate builds for CPython). If necessary, builds for PyPy Python can be re-enabled relatively easily, but does require a small numpy version change for `pp310-win_amd64`.

Linux aarch64 support is ~~currently a work in progress~~ **now working too!** Builds to take **much** longer though (~25 minutes because of aarch64 emulation on x86). 

**Reminders**


cometbeetle:
I believe the CI tests failed in version 3.12 because of `numpy 2.0.0`, which includes a few breaking changes. For now, I set `numpy<2.0.0` in `pyproject.toml`.

MarkWieczorek:
For linux, the problem seems to be 
```
Running before_all...
  
      + /opt/python/cp38-cp38/bin/python -c 'import sys, json, os; json.dump(os.environ.copy(), sys.stdout)'
      + sh -c 'yum install -y fftw-devel openblas-devel'
  Loaded plugins: fastestmirror, ovl
  Determining fastest mirrors
  Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
  14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
```
I don't know if that website is down, or if there is something we need to re-configure. I believe that this worked a couple of months ago...

cometbeetle:
I'm not sure. It runs on my fork though, so no worries if you can't approve things quickly!

MarkWieczorek:
On my system, everything works with numpy 2.0.0. Not sure why this is causing a problem for the CI.

cometbeetle:
I think it's because when the wheel is built with `oldest-supported-numpy`, the CI tests find that some functionality is not consistent between `numpy 1.x` and `numpy 2.x`. If `numpy` should only be supported > version 2.0.0, I can make that change. I just didn't know if you still want to guarantee support for older `numpy` versions.

MarkWieczorek:
I looked into numpy 2.0.0 a bit, and now I am more confused than before. I really have no idea what the best practice is here. If you think that pinning to <2.0.0 is the best solution, I'm fine with that.

cometbeetle:
I'll look into the numpy issue a bit more myself to see if can figure anything about. I think ideally it would be good to support the newer version if possible, but I'd want to make sure nothing breaks. I suppose the CI tests should tell us whether it's alright to do the upgrade. 

Also, I think I have Linux aarch64 builds working, however they take a long time because of emulation. I don't know how to test these builds, but I'm going to see if I can have them built concurrently with the standard manylinux builds (that way it doesn't take nearly as much total time).

MarkWieczorek:
For aarch64, I have an Apple computer running asahi linux, which should be compatible. If you tell me where to find the wheels, I can test it later tonight.

cometbeetle:
Let me know if you can see the progress of [my GitHub action](https://github.com/cometbeetle/SHTOOLS/actions/runs/9760425641). Once it's done, it should have aarch64 wheels available for download in a ZIP file.

cometbeetle:
So, it seems for now that trying to update `numpy` to 2.0.0 for the builds causes a number of issues that I don't know how to resolve. I think that while the kinks are being worked out of the dependencies (many of which use `numpy`), `numpy` should be pinned to `< 2.0.0`.

MarkWieczorek:
aarch64 works on Asahi Linux!

I'm going to look around for advice on how to deal with numpy 2.0. If I don't find anything, I'll just merge this at the end of the week. Thanks!

And since you are here.... Maybe you could look at this PR for the conda packaging: https://github.com/conda-forge/pyshtools-feedstock/pull/57 In summary, everything works, but in this PR, we tried to add aarch support and it broke something with the windows builds. I just haven't had the time to look into this, but if I put out a new release in a couple weeks, it would be nice to have aarch support for conda as well.

cometbeetle:
Great! I just merged in the changes to make aarch64 work for this PR. 

I'll take a look at the conda issue if I have some extra time. I also want to point out that building from source on Windows is extremely annoying for end-users right now, so really the only option for them on Windows will be to use the pre-compiled wheels.

cometbeetle:
I _think_ I figured_ out why numpy was causing issue. Apparently, there is a test that uses `matplotlib.cm.get_cmap`, which only works in matplotlib<3.9. For whatever reason, when numpy is updated, pip decides to download matplotlib 3.9.4, which raises an error. The good news is that matplotlib==3.8.4 supports python 3.12 **and** numpy, so we can just specify matplotlib<3.9 until we fix that test.

cometbeetle:
On second thought, let's just wait to deal with `numpy`. Updating to 2.0.0 requires too much adjustment of package versions, and I think it's the sort of upgrade that needs to be planned for. For example, `astropy `only supports `numpy 2.0.0` in `astropy>=6.1`, however this would require dropping support for Python 3.9. 

I'm going to leave the `numpy` issue alone for now.

MarkWieczorek:
I think I figured it out: https://numpy.org/devdocs/dev/depending_on_numpy.html#for-downstream-package-authors

in pyproject.toml, under `[project]`, add
```
  'numpy>=1.23.5',
```
as a dependency. Then under `build-system`, add
```
  'numpy>=2.0.0',
```
as a dependency.

In environment.yaml, we could add  `numpy>=1.23.5`.

I tried this on my old PR (without your changes) and it seems to work: https://github.com/SHTOOLS/SHTOOLS/pull/469

cometbeetle:
Interesting; that way it'll have support for `numpy 2.0.0`, but older versions will still work. 

Let me try to implement that later today, and if it works without any issues, I'll merge the changes into this PR.

MarkWieczorek:
Also, for the matplotlib issue, that is already fixed in the develop branch: https://github.com/SHTOOLS/SHTOOLS/pull/472

If it is really a problem for this PR, I would just rebase the PR onto the develop branch.

cometbeetle:
Good news: pulling the latest `develop` branch back into this PR fixed the `matplotlib` issue, and your `numpy` version pinning recommendations work great. 

I've now merged those changes into this PR.

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,139 @@
...
+        run: |
+          pipx run build --sdist
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cibw-sdist
+          path: dist/*.tar.gz
+
+  upload_pypi:
+    needs: [build_wheels, build_sdist]
```
MarkWieczorek, line 120: I think that we either need to add build_wheels_aarch64 here, or make a new job that just uploads the aarch64 wheels. Given that it takes longer build the later, I wonder if these should be uploaded separately.

QuLogic: It's better for consistency if they all upload at the same time, or else users will get mixed results depending on if they catch the small window of one done and not the other.

PR #498, opened 2024-12-05
in docs : fix inconsistencies in for csphase values

pphan-imcce:
Hi!

Noticed some inconsistencies in the documentation for either excluding or applying the csphase. Specifically:

- some parts of the docs say, respectively, (0) and (1)
- some say respectively (-1) and (1)
- and others say what I believe to be correct, respectively (1) and (-1). Please let me know if I misunderstood or missed anything.

Otherwise, the changes below should fix that.

(and minor change : the `epoch` entry was listed twice in the docs for [SHGravCoeffs class](https://shtools.github.io/SHTOOLS/python-shgravcoeffs.html))

Best regards,
PL

**Reminders**

- [ n/a] If adding new features, update the docstring to provide all information that is required to use the feature.


MarkWieczorek:
Thanks for catching this! I'll merge this, and I will put out a new release over the holidays. Eventually, I think that it would be better to use True and False for this parameter, and this might get changed in the next major version 5.0 that I am going to start working on next year.

Issue #510, opened 2025-04-23
Inconsistent handling of ducc backend when using optional lmax with MakeGridDHC


MarkWieczorek:
When creating a grid from complex coefficients using the `SHCoeffs.expand()` method, the ducc backend does not properly treat the optional `lmax` parameter. In particular, when `lmax` is greater than the maximum degree of the input coefficients, the coefficients should be zero padded so that the output grid has a finer sampling.

These examples show that everything works fine, except for the last line.

```
power = np.ones(11)

clm = pysh.SHCoeffs.from_random(power, kind='real')

clm.expand(backend='shtools').plot()
clm.expand(lmax=100, backend='shtools').plot()

clm.expand(backend='ducc').plot()
clm.expand(lmax=100, backend='ducc').plot()

clm2 = clm.convert(kind='complex')

clm2.expand(backend='shtools').plot()
clm2.expand(lmax=100, backend='shtools').plot()

clm2.expand(backend='ducc').plot()
clm2.expand(lmax=100, backend='ducc').plot()  # everything works except this!
```

And here is the error
```
File ~/SphericalHarmonics/shtools/pyshtools/backends/ducc0_wrapper.py:647, in MakeGridDHC(cilm, lmax, norm, sampling, csphase, lmax_calc, extend)
    645 lmax, lmax_calc, cilm = _prep_lmax(lmax, lmax_calc, cilm)
    646 alm = _ccilm2almi(cilm)
--> 647 alm = _apply_norm(alm, lmax, norm, csphase, False)
    648 res = _np.empty(
    649     [2 * lmax + 2 + extend, sampling * (2 * lmax + 2) + extend],
    650     dtype=_np.complex128,
    651 )
    652 _synthesize_DH(alm, lmax_calc, extend, res.imag)

File ~/SphericalHarmonics/shtools/pyshtools/backends/ducc0_wrapper.py:88, in _apply_norm(alm, lmax, norm, csphase, reverse)
     86 if reverse:
     87     lnorm = 1.0 / lnorm
---> 88 alm[0: lmax + 1] *= lnorm[0: lmax + 1]
     89 lnorm *= _np.sqrt(2.0) if reverse else (1.0 / _np.sqrt(2.0))
     90 mlnorm = -lnorm

ValueError: operands could not be broadcast together with shapes (66,) (101,) (66,)
```

@mreineck, would you mind taking a look at this?

mreineck:
Thanks, I'll have a look!

PR #511, opened 2025-04-23
Fix issue #510

mreineck:
This hopefully fixes the incorrect handling of `lmax` in the `MakeGrid` routines of the `ducc` backend.

**Reminders**


MarkWieczorek:
Looks good! I'll make the change myself for the shmagcoeffs and shgravcoeffs files.

mreineck:
To be honest, I'm not 100% sure about the change in `scoeffs.py`. The only important thing here is that `lmax_calc` will never be greater than `self.lmax`. A choice of `lmax_calc = min(self.lmax, lmax)` could also be valid.

MarkWieczorek:
I made the `lmax_calc = min(self.lmax, lmax)` change because this was causing problems with the examples. Thanks!

